PR #22, opened 2024-07-26
feat: agg-sender

goran-ethernal:
## Description

This PR implements the `Aggsender` component which is tasked of sending certificates to the `agglayer` using the `interop_sendCeritificate` rpc endpoint.

### Config
Next code is an example of the `Aggsender`  configuration:

```
[AggSender]
DBPath = "/tmp/aggsender" // path of the SQLite DB for the aggsender
AggLayerURL = "http://zkevm-agglayer" // URL to the agglayer
AggsenderPrivateKey = {Path = "/pk/sequencer.keystore", Password = "testonly"} // Aggsender private key that will sign the certificates before sending them to the agglayer
BlockGetInterval = "2s" // Interval for getting L1 blocks
URLRPCL2="http://test-aggoracle-l2:8545" // L2 node RPC URL
CheckSettledInterval = "2s" // Interval in which aggsender will check if sent certificates are settled
SaveCertificatesToFiles = false // indicates if sent certificates should be save to json files (for debugging purposes)
```

### Required components (services)

The `aggsender` utilizes next components in the cdk repo to build a certificate:

1. `L1 client` to get L1 blocks - certificates are settled on agglayer one by CDK in a single L1 epoch (configured by the `EpochSize` in the config), so we need to know how many L1 blocks have passed so we can send another certificate.
2. `L2 client` - to get the last block on the L2. This is needed to know to which block we get claims and bridges on L2 to send them in a certificate.
3. `L2 bridge syncer` - to get bridges, claims and blocks on L2, needed for a certificate.
4. `L1 info tree syncer` - to build inclusion proofs.
5. `Agglayer client` - needed to send certificates to the agglayer, and check if sent certificates were settled on the agglayer.

### DB

`Aggsender` saves sent certificates in its local `SQLite` db. It has one table called `certificate_info` in which it saves this data:

```go
type CertificateInfo struct {
	Height           uint64                     `meddler:"height"`
	CertificateID    common.Hash                `meddler:"certificate_id"`
	NewLocalExitRoot common.Hash                `meddler:"new_local_exit_root"`
	FromBlock        uint64                     `meddler:"from_block"`
	ToBlock          uint64                     `meddler:"to_block"`
	Status           agglayer.CertificateStatus `meddler:"status"`
}
```
`Height` - is a simple counter of this cdk's certificates. We can look at it as a nonce. It gets incremented by sending and settling certificates.
`CertificateID` - certificates Hash.
`NewLocalExitRoot `- new local exit root posted through the certificate (gotten from l2 bridge syncer).
`FromBlock` - block from which we got bridges and claims on L2.
`ToBlock` - block to which we got bridges and claims on L2.
`Status` - certificate status on `agglayer`. It can be `Pending` (pending to be settled), `Settled`, or `InError` (if some error happens on certificate settling).

### Flow

image


aggsender/aggsender.go
```diff
@@ -0,0 +1,314 @@
...
+		aggLayerClient:   aggLayerClient,
+		l1infoTreeSyncer: l1InfoTreeSyncer,
+		sequencerKey:     sequencerPrivateKey,
+		sendInterval:     cfg.CertificateSendInterval,
+	}, nil
+}
+
+// Start starts the AggSender
+func (a *AggSender) Start(ctx context.Context) {
+	go a.sendCertificates(ctx)
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/0add101f9452c09ed3e36baca5833199a1645533

aggsender/aggsender.go
```diff
@@ -0,0 +1,314 @@
...
+		GlobalIndex: &agglayer.GlobalIndex{
+			MainnetFlag: mainnetFlag,
+			RollupIndex: rollupIndex,
+			LeafIndex:   leafIndex,
+		},
+	}, nil
+}
+
+// sendCertificatesForNetwork sends certificates for a network
+func (a *AggSender) sendCertificatesForNetwork(ctx context.Context) error {
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/0add101f9452c09ed3e36baca5833199a1645533

aggsender/aggsender.go
```diff
@@ -0,0 +1,314 @@
...
+// buildCertificate builds a certificate from the bridge events
+func (a *AggSender) buildCertificate(ctx context.Context,
+	bridges []bridgesync.Bridge,
+	claims []bridgesync.Claim,
+	previousExitRoot common.Hash, lastHeight uint64) (*agglayer.Certificate, error) {
+	bridgeExits := make([]*agglayer.BridgeExit, 0, len(bridges))
+	importedBridgeExits := make([]*agglayer.ImportedBridgeExit, 0, len(claims))
+
+	for _, bridge := range bridges {
+		bridgeExit := convertBridgeToBridgeExit(bridge)
```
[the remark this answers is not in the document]
goran-ethernal: I know, I was just throwing code, since it is being changed on regular basis by the agglayer team. In the newest commit: https://github.com/0xPolygon/cdk/pull/22/commits/0add101f9452c09ed3e36baca5833199a1645533, you will see that they changed the API structure quite a bit, especially in regards to the `L1InfoTreeLeaves` and `MerkleProofs`, so I don't think we can change the `bridgesync` to work with agglayer API types. I mean, we could, but it won't have all the data, so I am not a fan of this approach. 

The second approach of custom marshaller is also not a good idea, because of the data we collect after we get bridges and claims from the `bridgesync`, the proofs, the global exit root, block timestamp, block hash, etc.

I am not sure what would be the right approach. Maybe changing the `bridgesync` would be the easiest and the correct approach.

aggsender/aggsender.go
```diff
@@ -0,0 +1,314 @@
...
+
+	if err := a.aggLayerClient.SendCertificate(signedCertificate); err != nil {
+		return fmt.Errorf("error sending certificate: %w", err)
+	}
+
+	if err := a.saveLastSentCertificate(ctx, lastFinalizedBlock.Nonce.Uint64(), certificate); err != nil {
+		return fmt.Errorf("error saving last sent certificate in db: %w", err)
+	}
+
+	return nil
```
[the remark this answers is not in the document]
goran-ethernal: Indeed. We will remove this database code in `aggsender` after they introduce the new endpoint to get the latest submitted certificate, as well as the change in the current one.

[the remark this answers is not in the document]
goran-ethernal: Implemented the `checkIfCertificatesAreSettled` function for this, based on the new endpoint on the `agglayer` called, `interop_getCertificateHeader`.

aggsender/aggsender.go
```diff
@@ -0,0 +1,495 @@
...
+
+// shouldSendCertificate checks if a certificate should be sent at given L1 block
+// we send certificates at two blocks before the epoch ending so we get most of the
+// bridges and claims in that epoch
+func (a *AggSender) shouldSendCertificate(block uint64) bool {
+	if block == 0 {
+		return false
+	}
+
+	return (block+numOfBlocksBeforeEpochEnding)%a.cfg.EpochSize == 0
```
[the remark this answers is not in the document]
goran-ethernal: I agree, I am changing this currently, to not have this send interval, rather just pinging blocks, until we reach the desired one.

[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/8da0c737bd9dd4054d48ef97caa09fd4f77290f0

agglayer/types.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// Hash returns a hash that uniquely identifies the certificate
+func (c *Certificate) Hash() common.Hash {
+	return crypto.Keccak256Hash(
+		cdkcommon.Uint32ToBytes(c.NetworkID),
+		cdkcommon.Uint64ToBytes(c.Height),
+		c.PrevLocalExitRoot.Bytes(),
+		c.NewLocalExitRoot.Bytes(),
+	)
```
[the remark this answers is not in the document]
goran-ethernal: They were before, but currently, they are not. These are left because of the old agglayer spec. I can remove it since they are not used.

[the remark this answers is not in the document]
goran-ethernal: I see now they've had changes on the main branch that might use these functions, so I will leave them until we update again.

agglayer/types.go
```diff
@@ -0,0 +1,182 @@
...
+	LeafType           LeafType       `json:"leaf_type"`
+	TokenInfo          *TokenInfo     `json:"token_info"`
+	DestinationNetwork uint32         `json:"destination_network"`
+	DestinationAddress common.Address `json:"destination_address"`
+	Amount             *big.Int       `json:"amount"`
+	Metadata           []byte         `json:"metadata"`
+}
+
+// Hash returns a hash that uniquely identifies the bridge exit
+func (c *BridgeExit) Hash() common.Hash {
```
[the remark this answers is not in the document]
goran-ethernal: I will leave them, since there are changes on the agglayer main branch that use these functions, so we will need to update in the near future.

aggsender/mock_eth_client.go
```diff
@@ -0,0 +1,89 @@
+// Code generated by mockery v2.45.0. DO NOT EDIT.
```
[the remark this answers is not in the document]
goran-ethernal: We don't have a clear standard on this. Some packages have it in mocks folder, some don't. I will move it to mocks, but It is not something that is forced.

agglayer/client.go
```diff
@@ -21,6 +21,8 @@ var ErrAgglayerRateLimitExceeded = fmt.Errorf("agglayer rate limit exceeded")
 type AgglayerClientInterface interface {
 	SendTx(signedTx SignedTx) (common.Hash, error)
 	WaitTxToBeMined(hash common.Hash, ctx context.Context) error
+	SendCertificate(certificate *SignedCertificate) (common.Hash, error)
+	GetCertificateHeader(certificateHash common.Hash) (*CertificateHeader, error)
```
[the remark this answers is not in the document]
goran-ethernal: We don't need that. Certificates are sent 1 per epoch, and epochs will be a reasonable size of blocks. Enough time to settle the sent certificate (or put it in InError), before we send another one.

aggsender/db/migrations/migrations.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	_ "embed"
+
+	"github.com/0xPolygon/cdk/db"
+	"github.com/0xPolygon/cdk/db/types"
+)
+
+//go:embed 0001.sql
+var mig001 string
```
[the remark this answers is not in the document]
goran-ethernal: I would leave that for another PR, it requires a lot of changes.

agglayer/types.go
```diff
@@ -0,0 +1,182 @@
...
+
+type LeafType uint8
+
+func (l LeafType) Uint8() uint8 {
+	return uint8(l)
+}
+
+const (
+	LeafTypeAsset   LeafType = 0
+	LeafTypeMessage LeafType = 1
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/bdad93de39689ed43a97159102a10cbcf89b7f7c

aggsender/types/types.go
```diff
@@ -0,0 +1,21 @@
...
+	Height           uint64                     `meddler:"height"`
+	CertificateID    common.Hash                `meddler:"certificate_id"`
+	NewLocalExitRoot common.Hash                `meddler:"new_local_exit_root"`
+	FromBlock        uint64                     `meddler:"from_block"`
+	ToBlock          uint64                     `meddler:"to_block"`
+	Status           agglayer.CertificateStatus `meddler:"status"`
+}
+
+func (c CertificateInfo) String() string {
+	return fmt.Sprintf("Height: %d, CertificateID: %s", c.Height, c.CertificateID.String())
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/bdad93de39689ed43a97159102a10cbcf89b7f7c

aggsender/config.go
```diff
@@ -0,0 +1,16 @@
+package aggsender
+
+import (
+	"github.com/0xPolygon/cdk/config/types"
+)
+
+// Config is the configuration for the AggSender
+type Config struct {
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/bdad93de39689ed43a97159102a10cbcf89b7f7c

aggsender/aggsender.go
```diff
@@ -0,0 +1,523 @@
...
+	"github.com/0xPolygon/cdk/l1infotreesync"
+	"github.com/0xPolygon/cdk/log"
+	treeTypes "github.com/0xPolygon/cdk/tree/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// defines number of blocks before epoch ending to send a certificate
+const numOfBlocksBeforeEpochEnding = 2
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/bdad93de39689ed43a97159102a10cbcf89b7f7c

aggsender/aggsender.go
```diff
@@ -0,0 +1,523 @@
...
+			return
+		}
+	}
+}
+
+// sendCertificate sends certificate for a network
+func (a *AggSender) sendCertificate(ctx context.Context) error {
+	a.log.Infof("trying to send a new certificate...")
+
+	block, err := a.l1Client.BlockNumber(ctx)
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/bdad93de39689ed43a97159102a10cbcf89b7f7c

aggsender/aggsender.go
```diff
@@ -0,0 +1,523 @@
...
+		Height:           certificate.Height,
+		CertificateID:    certificateHash,
+		NewLocalExitRoot: certificate.NewLocalExitRoot,
+		FromBlock:        fromBlock,
+		ToBlock:          toBlock,
+	}); err != nil {
+		return fmt.Errorf("error saving last sent certificate in db: %w", err)
+	}
+
+	a.log.Infof("certificate: %s sent successfully for block: %d to block: %d", certificateHash, fromBlock, toBlock)
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/bdad93de39689ed43a97159102a10cbcf89b7f7c

agglayer/types.go
```diff
@@ -0,0 +1,182 @@
...
+	Height           uint64            `json:"height"`
+	EpochNumber      *uint64           `json:"epoch_number"`
+	CertificateIndex *uint64           `json:"certificate_index"`
+	CertificateID    common.Hash       `json:"certificate_id"`
+	NewLocalExitRoot common.Hash       `json:"new_local_exit_root"`
+	Status           CertificateStatus `json:"status"`
+}
+
+func (c CertificateHeader) String() string {
+	return fmt.Sprintf("Height: %d, CertificateID: %s", c.Height, c.CertificateID.String())
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/bdad93de39689ed43a97159102a10cbcf89b7f7c

aggsender/config.go
```diff
@@ -0,0 +1,16 @@
...
+	DBPath               string                   `mapstructure:"DBPath"`
+	AggLayerURL          string                   `mapstructure:"AggLayerURL"`
+	BlockGetInterval     types.Duration           `mapstructure:"BlockGetInterval"`
+	CheckSettledInterval types.Duration           `mapstructure:"CheckSettledInterval"`
+	SequencerPrivateKey  types.KeystoreFileConfig `mapstructure:"SequencerPrivateKey"`
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/bdad93de39689ed43a97159102a10cbcf89b7f7c

agglayer/client.go
```diff
@@ -86,3 +88,43 @@ func (c *AggLayerClient) WaitTxToBeMined(hash common.Hash, ctx context.Context)
...
+
+// SendCertificate sends a certificate to the AggLayer
+func (c *AggLayerClient) SendCertificate(certificate *SignedCertificate) (common.Hash, error) {
+	response, err := rpc.JSONRPCCall(c.url, "interop_sendCertificate", certificate)
+	if err != nil {
+		return common.Hash{}, err
+	}
+
+	if response.Error != nil {
+		return common.Hash{}, fmt.Errorf("%v %v", response.Error.Code, response.Error.Message)
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/bdad93de39689ed43a97159102a10cbcf89b7f7c

agglayer/client.go
```diff
@@ -86,3 +88,43 @@ func (c *AggLayerClient) WaitTxToBeMined(hash common.Hash, ctx context.Context)
...
+
+// GetCertificateHeader returns the certificate header associated to the hash
+func (c *AggLayerClient) GetCertificateHeader(certificateHash common.Hash) (*CertificateHeader, error) {
+	response, err := rpc.JSONRPCCall(c.url, "interop_getCertificateHeader", certificateHash)
+	if err != nil {
+		return nil, err
+	}
+
+	if response.Error != nil {
+		return nil, fmt.Errorf("%v %v", response.Error.Code, response.Error.Message)
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/bdad93de39689ed43a97159102a10cbcf89b7f7c

aggsender/mock_eth_client.go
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/313cf7fccc42354520a70b5bdddaff22374d41b5

joanestebanr (approved):

Stefan-Ethernal (approved):

aggsender/aggsender.go
```diff
@@ -0,0 +1,523 @@
...
+)
+
+var errNoBridgesAndClaims = errors.New("no bridges and claims to build certificate")
+
+// AggSender is a component that will send certificates to the aggLayer
+type AggSender struct {
+	log aggsendertypes.Logger
+
+	l2Syncer         aggsendertypes.L2BridgeSyncer
+	l2Client         aggsendertypes.EthClient
```
arnaubennassar, line 30: we should avoid using a execution client in the aggSender. Doing so will couple the aggSender with the execution client, something that we want to avoid to keep thing modular. Basically having EVM stuff on the aggsender will complicate the integration of non-EVM stuff.

It looks like that removing this should be easy, if we relay on the last synced block for chosing the "to block"

aggsender/aggsender.go
```diff
@@ -0,0 +1,523 @@
...
+	fromBlock := lastCertificateBlock + 1
+	toBlock, err := chooseToL2Block(lastL2Block.Number.Uint64(), lasL2BlockSynced, lastCertificateBlock)
+	if err != nil {
+		a.log.Errorf("no new blocks to send a certificate: %s", err.Error())
+		return nil
+	}
+	a.log.Infof("lastL2Block: %d, lasL2BlockSynced: %d  choose range:[%d,%d]",
+		lastL2Block.Number.Uint64(), lasL2BlockSynced, fromBlock, toBlock)
+
+	bridges, err := a.l2Syncer.GetBridges(ctx, fromBlock, toBlock)
```
arnaubennassar, line 138: not sure how the len() == 0 is returned, maybe it's with the form of an error not found

aggsender/aggsender.go
```diff
@@ -0,0 +1,523 @@
...
+	if err != nil {
+		return fmt.Errorf("error getting bridges: %w", err)
+	}
+
+	if len(bridges) == 0 {
+		a.log.Infof("no bridges consumed, no need to send a certificate from block: %d to block: %d", fromBlock, toBlock)
+		return nil
+	}
+
+	claims, err := a.l2Syncer.GetClaims(ctx, fromBlock, toBlock)
```
arnaubennassar, line 148: same comment as the potential error not found as before, in this case that would be bad, as it's acceptable to not have claims and so we should proceed with the execution

aggsender/aggsender.go
```diff
@@ -0,0 +1,523 @@
...
+
+	var (
+		previousLocalExitRoot common.Hash
+		previousHeight        uint64
+		lastCertificateBlock  uint64
+	)
+
+	if lastSentCertificate.CertificateID != (common.Hash{}) {
+		// we have sent a certificate before, get the last certificate header
+		lastSentCertificateHeader, err := a.aggLayerClient.GetCertificateHeader(lastSentCertificate.CertificateID)
```
arnaubennassar, line 260: I dont get this part, I thought it would be something like:
- if we have a last sent cert on the DB, return that (prolly after asserting that what we have is consistent with the aggLayer)
- else, get it from the aggLayer

But looks like if there is nothing on the DB we assume that there is nothing on the aggLayer, which may not be the case...

aggsender/aggsender.go
```diff
@@ -0,0 +1,523 @@
...
+	a.lastL1CertificateBlock = l1Block
+	a.lock.Unlock()
+
+	a.log.Infof("certificate: %s sent successfully for range of l2 blocks (from block: %d, to block: %d)",
+		certificateHash, fromBlock, toBlock)
+
+	return nil
+}
+
+func chooseToL2Block(lastL2Block, lasL2BlockSynced, lastCertificateBlock uint64) (uint64, error) {
```
arnaubennassar, line 190: If I understood things well, wen we do the `getLastSentCertificateData` we relay on the LER for linking to a block number. So I guess that we should be consistent here. Otherwise we could send claims twice. Example:

- last block sent is 10
- block 18 includes a bridge => so a new LER
- block 20 includes a claim
- last block is 22

aggSender sends from 11 to 22

next iteration aggSender picks block 18 as last block sent as it matches with the last LER sent, and so it sends again the claim from block 20

aggsender/aggsender.go
```diff
@@ -0,0 +1,523 @@
...
+				continue
+			}
+		}
+	}
+}
+
+// shouldSendCertificate checks if a certificate should be sent at given L1 block
+// we send certificates at two blocks before the epoch ending so we get most of the
+// bridges and claims in that epoch
+func (a *AggSender) shouldSendCertificate(block uint64) bool {
```
arnaubennassar, line 511: prolly a good idea to check if we've sent some cert that is still pending and return false if so

aggsender/db/migrations/0001.sql
```diff
@@ -0,0 +1,14 @@
+-- +migrate Down
+DROP TABLE IF EXISTS certificate_info;
+
+-- +migrate Up
+CREATE TABLE certificate_info (
+	height                      INTEGER NOT NULL,
+	certificate_id              VARCHAR NOT NULL,
```
arnaubennassar, line 7: fix identation

aggsender/db/migrations/0001.sql
```diff
@@ -0,0 +1,14 @@
...
+    status                      INTEGER NOT NULL,
+    new_local_exit_root         VARCHAR NOT NULL,
+    from_block                  INTEGER NOT NULL,
+    to_block                    INTEGER NOT NULL,
+
+    PRIMARY KEY (height, certificate_id)
```
arnaubennassar, line 13: shouldnt `certificate_id` be enough for a primary key?

bridgesync/migrations/bridgesync0001.sql
```diff
@@ -25,11 +25,11 @@ CREATE TABLE bridge (
 CREATE TABLE claim (
 	block_num               INTEGER NOT NULL REFERENCES block(num) ON DELETE CASCADE,
 	block_pos               INTEGER NOT NULL,
-	global_index            DECIMAL(78, 0) NOT NULL,
+	global_index           	TEXT NOT NULL,
```
arnaubennassar, line 28: why this change?

arnaubennassar (changes requested):

aggsender/aggsender.go
goran-ethernal: When you've processed blocks but don't have any bridges in them.

aggsender/aggsender.go
goran-ethernal: We will continue building certificate if we don't have claims (`len() == 0`). We are protecting our selves from block not processed in `toBlock, err := chooseToL2Block(lastL2Block.Number.Uint64(), lasL2BlockSynced, lastCertificateBlock)`.

aggsender/aggsender.go
goran-ethernal: I agree, but the only endpoint the `agglayer` has in `getCertificateHeader` which gets the certificate by its hash. If we don't have anything in our db, then there is no way to know what was last settled on the `agglayer`. We need to raise this with them.

bridgesync/migrations/bridgesync0001.sql
goran-ethernal: The change on the bridge syncer db to store big int as TEXT was made because the sqlite kept saving big numbers with scientific annotation, for example if you have 810.000.000 it will save it as 8.1e9, and when reading it back to big int, big int throws an error.




NEW

[11:27](https://ethernalworkspace.slack.com/archives/D06H6U9QU9Z/p1729718857298149)
so we need to save it as TEXT so that sqlite dont do this magic with numbers and save them with exponents

aggsender/aggsender.go
joanestebanr: The `fromBlock` is the block of last certificate + 1, so the case that you say the next certificate will start with **22+1**

aggsender/aggsender.go
goran-ethernal: Removed execution clients from aggsender: https://github.com/0xPolygon/cdk/pull/22/commits/df721ecca5dd141e832b1cd31e433bf43e6232db

aggsender/aggsender.go
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/df721ecca5dd141e832b1cd31e433bf43e6232db

aggsender/db/migrations/0001.sql
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/21680d828f80c61e923a2ce5d340d00c02dc3088

aggsender/db/migrations/0001.sql
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/21680d828f80c61e923a2ce5d340d00c02dc3088

aggsender/aggsender.go
goran-ethernal: Arnau is right, because in the case when we have previously sent a certificate, we will get the last block based on previous sent LER: `lastCertificateBlock, err = a.l2Syncer.GetBlockByLER(ctx, previousLocalExitRoot)`.

I'll change this so we return the `lastSentCertificate.ToBlock`.

goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/d1e805ab5c3c288cc34814072c66e19732028c9f

aggsender/aggsender.go
goran-ethernal: We are talking with the agglayer team, so they will probably add a new rpc point for this.

aggsender/aggsender.go
```diff
@@ -0,0 +1,466 @@
...
+
+	if lastSentCertificate.CertificateID != (common.Hash{}) {
+		// we have sent a certificate before, get the last certificate header
+		lastSentCertificateHeader, err := a.aggLayerClient.GetCertificateHeader(lastSentCertificate.CertificateID)
+		if err != nil {
+			return common.Hash{}, 0, 0, fmt.Errorf("error getting certificate %s header: %w",
+				lastSentCertificate.CertificateID, err)
+		}
+
+		if lastSentCertificateHeader.Status == agglayer.InError {
```
joanestebanr, line 230: it looks like in that case you don't set properly `lastCertificateBlock`

goran-ethernal: Fixed: https://github.com/0xPolygon/cdk/pull/22/commits/335b15683579f711848d5c3f51de1ab7f26a3a75

arnaubennassar (approved):

aggsender/config.go
```diff
@@ -0,0 +1,21 @@
+package aggsender
+
+import (
+	"github.com/0xPolygon/cdk/config/types"
+)
+
+// Config is the configuration for the AggSender
+type Config struct {
+	// DBPath is the path of the sqlite db on which the AggSender will store the data
+	DBPath string `mapstructure:"DBPath"`
```
vcastellm, line 10: This could be renamed to StoragePath for consistency with other components, we can use this PR for that maybe?

bridgesync/migrations/bridgesync0001.sql
vcastellm: Weird... is it a benefit to store this in strings and do the parsing on the app side? I'd say it's better to do this https://stackoverflow.com/a/40435685

bridgesync/migrations/bridgesync0001.sql
```diff
@@ -25,11 +25,11 @@ CREATE TABLE bridge (
...
 	origin_network          INTEGER NOT NULL,
 	origin_address          VARCHAR NOT NULL,
 	destination_address     VARCHAR NOT NULL,
-	amount                  DECIMAL(78, 0) NOT NULL,
+	amount                  TEXT NOT NULL,
```
vcastellm, line 32: same for this

vcastellm (changes requested):
Observe some comments

bridgesync/migrations/bridgesync0001.sql
vcastellm: INTs in SQLite are 64 bit size so they are the same supported size as Postgres BIGINT, better to switch to INT?

bridgesync/bridge_contract_test.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLastUpdatedDepositCount(t *testing.T) {
+	_, err := NewBridgeContract(common.Address{}, nil)
```
vcastellm, line 11: Is this test necessary?

cmd/run.go
```diff
@@ -12,10 +12,12 @@ import (
 	zkevm "github.com/0xPolygon/cdk"
 	dataCommitteeClient "github.com/0xPolygon/cdk-data-availability/client"
 	jRPC "github.com/0xPolygon/cdk-rpc/rpc"
+	"github.com/0xPolygon/cdk/agglayer"
 	"github.com/0xPolygon/cdk/aggoracle"
 	"github.com/0xPolygon/cdk/aggoracle/chaingersender"
 	"github.com/0xPolygon/cdk/aggregator"
 	"github.com/0xPolygon/cdk/aggregator/db"
+	"github.com/0xPolygon/cdk/aggsender"
```
vcastellm, line 20: duplicated?

common/components.go
```diff
@@ -13,4 +13,6 @@ const (
 	CLAIM_SPONSOR = "claim-sponsor" //nolint:stylecheck
 	// PROVER name to identify the prover component
 	PROVER = "prover"
```
vcastellm, line 15: not related to this PR but prover here?

config/default.go
```diff
@@ -7,7 +7,7 @@ L1URL = "http://localhost:8545"
 L2URL = "localhost:8123"
 L1AggOracleURL = "http://test-aggoracle-l1:8545"
 L2AggOracleURL = "http://test-aggoracle-l2:8545"
-
+AggLayerURL = "https://agglayer-dev.polygon.technology"
```
vcastellm, line 10: Can I run without the agglayer? in other words, is this conditional with setting `SettlementBackend` in the aggregator?

scripts/local_config
```diff
@@ -30,10 +30,13 @@ function get_value_from_toml_file(){
     local _LINE
     local _inside_section=0
     local _return_next_line=0
+    local _TMP_FILE=$(mktemp)
+    cat $_FILE > $_TMP_FILE
+    # Maybe the file doesnt end with a new line so we added just in case
```
vcastellm, line 35: And this is why files should always end in new line 😁

scripts/local_config
```diff
@@ -30,10 +30,13 @@ function get_value_from_toml_file(){
...
+    echo " " >> $_TMP_FILE
```
vcastellm, line 36: This should be enough

```suggestion
    echo >> $_TMP_FILE
```

aggsender/config.go
joanestebanr: In fact, most of the  components used `DBPath`:bridgesync, claimsponsor, reorgdetector, l1infotreesync,....

common/components.go
joanestebanr: You're right  that is not a compoment, but I don't known if this is the best PR to fix this

config/default.go
joanestebanr: Sadly we don't have conditional config yet

test/helpers/lxly-bridge-test.bash
```diff
@@ -37,6 +37,7 @@ function claim() {
     readonly claimable_deposit_file=$(mktemp)
     echo "Getting full list of deposits" >&3
     curl -s "$bridge_api_url/bridges/$destination_addr?limit=100&offset=0" | jq '.' | tee $bridge_deposit_file
+    cat $bridge_deposit_file >&3
```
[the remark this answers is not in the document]
goran-ethernal: I am not sure what this line is. Might have been a leftover of some rebase.

aggsender/config.go
vcastellm: Yes but after that naming we decided naming it `StoragePath` in ethtxmanager see https://github.com/0xPolygon/cdk/blob/95cdb6afe1bbb14e5cf8e6f5e1fad3ac3c2681a6/config/default.go#L56 and the agreement was to update progressively the rest, why? Because it's called a `Storage` type not `DB` https://github.com/0xPolygon/cdk/pull/22/files#diff-e381a7a01497b1c89fcd51acf63d7f51713a9b25c747ed857fb0ba12b2bfca98R22 so for the sake of consistency it should be called Storage.

cmd/run.go
vcastellm: My mistake

test/helpers/lxly-bridge-test.bash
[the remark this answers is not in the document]
vcastellm: Remove?

bridgesync/migrations/bridgesync0001.sql
vcastellm: @goran-ethernal not sure why it's not showing in the comments, but this is important

vcastellm:
This is being kept hidden by the GithubUI but it's important https://github.com/0xPolygon/cdk/pull/22#discussion_r1816496554

bridgesync/migrations/bridgesync0001.sql
goran-ethernal: I mean we can, but I don't see a benefit. In text we can save it as a hash, or hex.

aggsender/config.go
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/22/commits/73188555ba2181303c9936c085fc0921997f7721

arnaubennassar (dismissed):
plz revert the changes done in https://github.com/0xPolygon/cdk/pull/140 before merging

joanestebanr:
> plz revert the changes done in #140 before merging

If we revert this PR we are going to have a bug if somebody calls `bridgeAsset` of Bridge Contract with param `forceUpdateGlobalExitRoot=false`. 
I suggest:
- Keep this code until we have a better solution (think in that as a workarround)   until contracts are updated as you suggest
- If we decide to delete this code we can implement a less efficient solution: 
    - add a check to the synced Bridge and if have `forceUpdateGlobalExitRoot=false` prevent syncing  this and keep in a 'waiting assets' lists on DB until appears a new  `bridgeAsset` with `forceUpdateGlobalExitRoot=true` that then we can include the 'waiting assets list' .This suppose that  in that in some moment there are going to be this call because the call `updateGlobalExitRoot()`that is supported in #140 solution is no longer supported.
 - Don't implement any solution to the bug and just detect it showing an error  and stopping syncing

bridgesync/migrations/bridgesync0001.sql
vcastellm: well, there's some differences, int is not the same as bytes, less space, more performant, why you want to save an hex or hash in this field?

bridgesync/migrations/bridgesync0001.sql
```diff
@@ -16,7 +16,7 @@ CREATE TABLE bridge (
 	origin_address      VARCHAR NOT NULL,
 	destination_network INTEGER NOT NULL,
 	destination_address VARCHAR NOT NULL,
-	amount              DECIMAL(78, 0) NOT NULL,
+	amount              TEXT NOT NULL,
```
vcastellm, line 19: Same here, it can be INT, more performance less conversions on app side, less used space, direct mapping data model

bridgesync/migrations/bridgesync0001.sql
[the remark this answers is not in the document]
vcastellm: Same here, it can be INT, more performance less conversions on app side, less used space, direct mapping data model, etc. Let's check if that works

bridgesync/migrations/bridgesync0001.sql
goran-ethernal: BigInt in Golang can be more than 64 bits. To humor you, I've set amount and global_index to integers and made a UT that stores a claim that has a large big.Int value in the global_index. It again stores the value with scientific annotation, and when we try to set the `big.Int.SetString` it fails:

```
meddler.WriteTargets: PostRead error on column [global_index]: big.Int.SetString failed on "1.8446744073709552e+19"
```

goran-ethernal: I've pushed the test: https://github.com/0xPolygon/cdk/pull/22/commits/12ca5ef16aaf1b04b49a177c9983b628eb5e4c33. You can try it yourself. Just change the column type in .sql file.

vcastellm: this is interesting, how do we would do in postgres? bigint or string? same here, if we need Go big ints then it's string if we are ok with Postgres Bigints then Ints here

goran-ethernal: I really don't know. Can't find any example currently on cdk repo, where we read something in postgress to a big.Int.

bridgesync/migrations/bridgesync0001.sql
vcastellm: in this case, the global index is probably enough with SQLite INTs but it's not worth it

bridgesync/migrations/bridgesync0001.sql
vcastellm: Ok, I checked it, wei can outgrow this type of field, so let's go with text.

vcastellm (approved):

PR #75, opened 2024-09-16
test: Added check for EOA transfer E2E Test

rachit77:
## Description

Added additional balance check for ether transfers


test/helpers/common.bash
```diff
@@ -185,3 +198,59 @@ function queryContract() {
...
+
+    # Check if rpc_url is available
+    if [[ -z "$rpc_url" ]]; then
+        echo "Error: rpc_url environment variable is not set."
+        return 1
+    fi
+
+    # Use cast to perform a generic RPC call
+    local response
+    response=$(cast --rpc-url "$rpc_url" "$method" "${params[@]}" 2>&1)
```
Stefan-Ethernal, line 215: The usage of rpcQuery function is generally incorrect:
- https://github.com/0xPolygon/cdk/pull/75/files#diff-804cd8944aa39afffc98754304b5107919eb640132a5e4f9c63a0ac403b5d6d4R25, https://github.com/0xPolygon/cdk/pull/75/files#diff-0adaebba413a6e19d5601c96ee3d7af4d70464bad65d781b468aee6ea348f69bR236, https://github.com/0xPolygon/cdk/pull/75/files#diff-0adaebba413a6e19d5601c96ee3d7af4d70464bad65d781b468aee6ea348f69bR246 etc. - for function name, provide `eth_getBalance` and provide two parameters: address and latest

Example: 
```bash
cast rpc --rpc-url http://127.0.0.1:32953 eth_getBalance 0x85dA99c8a7C2C95964c8EfD687E95E632Fc533D6 latest
```
- https://github.com/0xPolygon/cdk/pull/75/files#diff-804cd8944aa39afffc98754304b5107919eb640132a5e4f9c63a0ac403b5d6d4R14 - provide `eth_getTransactionCount` as the function name and account address and latest as the param values

Example:
```bash
cast rpc --rpc-url http://127.0.0.1:32953 eth_getTransactionCount 0x85dA99c8a7C2C95964c8EfD687E95E632Fc533D6 latest
```

```suggestion
    response=$(cast rpc --rpc-url "$rpc_url" "$method" "${params[@]}" 2>&1)
```

If we go with your approach, that would work as well, but this way we can invoke arbitrary JSON RPC function (instead of just a couple of in-built `cast balance`, `cast nonce` etc. commands), which gives us greater flexibility. Also bear in mind that with the initial approach you took, subcommand name must go right after the `cast` command (and not after the `--rpc-url` flag).

test/helpers/common.bash
```diff
@@ -185,3 +199,58 @@ function queryContract() {
...
+
+    # Check if rpc_url is available
+    if [[ -z "$rpc_url" ]]; then
+        echo "Error: rpc_url environment variable is not set."
+        return 1
+    fi
+
+    # Use cast to perform a generic RPC call
+    local response
+    response=$(cast rpc "$method" "${params[@]}" --rpc-url "$rpc_url" 2>&1)
```
Stefan-Ethernal, line 216: I'm suggesting this, as there might be multiple parameters:

```suggestion
    response=$(cast rpc --rpc-url "$rpc_url" "$method" "${params[@]}"  2>&1)
```

test/helpers/common.bash
```diff
@@ -185,3 +195,67 @@ function queryContract() {
 
     return 0
 }
+
+function rpcQuery() {
```
Stefan-Ethernal, line 199: This function seems unused, so is it safe to delete it?

test/basic-e2e.bats
```diff
@@ -11,11 +11,26 @@ setup() {
 }
 
 @test "Send EOA transaction" {
+    local sender_addr=$(cast wallet address --private-key $private_key)
+    local initial_nonce=$(cast nonce "$sender_addr" --rpc-url "$rpc_url") || return 1
     local value="10ether"
 
+    # case 1: Transaction successfull sender has sufficient balance
```
Stefan-Ethernal, line 18: Typo
```suggestion
    # case 1: Transaction successful sender has sufficient balance
```

test/basic-e2e.bats
```diff
@@ -11,11 +11,26 @@ setup() {
...
     run sendTx "$private_key" "$receiver" "$value"
     assert_success
     assert_output --regexp "Transaction successful \(transaction hash: 0x[a-fA-F0-9]{64}\)"
+
+    # case 2: Transaction rejected sender attempts to transfer more than they have in their wallet
```
Stefan-Ethernal, line 23:
```suggestion
    # case 2: Transaction rejected as sender attempts to transfer more than it has in its wallet.
```

test/basic-e2e.bats
```diff
@@ -11,11 +11,26 @@ setup() {
...
+    # Transaction will fail pre-validation check on the node and will be dropped subsequently from the pool
+    # without recording it on the chain and hence nonce will not change
+    local sender_balance=$(cast balance "$sender_addr" --ether --rpc-url "$rpc_url") || return 1    
+    local excessive_value=$(echo "$sender_balance + 1" | bc)"ether"
+    run sendTx "$private_key" "$receiver" "$excessive_value"
+    assert_failure
+
+    # check whether nonce of sender was updated correctly
```
Stefan-Ethernal, line 31:
```suggestion
    # Check whether the sender's nonce was updated correctly
```

test/helpers/common.bash
```diff
@@ -185,3 +195,67 @@ function queryContract() {
...
+        echo "Error: Failed to perform RPC query."
+        echo "$response"
+        return 1
+    fi
+
+    echo "$response"
+    return 0
+}
+
+function checkTransactionSuccess() {
```
Stefan-Ethernal, line 225:
```suggestion
function checkBalances() {
```

Also, consider adding the validation of the input parameters:
- addresses should match the `0x[a-fA-F0-9]{40}` regex
- transaction hash should match the `0x[a-fA-F0-9]{64}` regex
etc.

test/helpers/common.bash
```diff
@@ -144,6 +146,14 @@ function sendTx() {
         return 1
     fi
 
+    if [[ "$value_or_function_sig" =~ ^[0-9]+(ether)?$ ]]; then
+        checkTransactionSuccess "$sender_addr" "$receiver" "$value_or_function_sig" "$tx_hash" "$sender_initial_balance" "$receiver_initial_balance"
+        if [[ $? -ne 0 ]]; then
+            echo "Error: Balance not updated correctly."
+            return 1
+        fi
```
Stefan-Ethernal, line 154: Why don't you move this function call here in the if that already checks whether it is a value transfer (https://github.com/0xPolygon/cdk/pull/75/files#diff-0adaebba413a6e19d5601c96ee3d7af4d70464bad65d781b468aee6ea348f69bR111)?

test/helpers/common.bash
```diff
@@ -185,3 +195,67 @@ function queryContract() {
...
+    fi
+
+    echo "$response"
+    return 0
+}
+
+function checkTransactionSuccess() {
+    local sender_addr="$1"
+    local receiver="$2"
+    local value_or_function_sig="$3"
```
Stefan-Ethernal, line 228: Why not simply call it `amount` or `value`?
```suggestion
    local amount="$3"
```

test/helpers/common.bash
```diff
@@ -185,3 +195,67 @@ function queryContract() {
...
+        echo "$response"
+        return 1
+    fi
+
+    echo "$response"
+    return 0
+}
+
+function checkTransactionSuccess() {
+    local sender_addr="$1"
```
Stefan-Ethernal, line 226: Call it `sender` (consistent with the `receiver` parameter). WDYT?
```suggestion
    local sender="$1"
```

test/basic-e2e.bats
```diff
@@ -11,11 +11,26 @@ setup() {
 }
 
 @test "Send EOA transaction" {
+    local sender_addr=$(cast wallet address --private-key $private_key)
```
Stefan-Ethernal, line 14:
```suggestion
    local sender_addr=$(cast wallet address --private-key "$private_key)"
```

Stefan-Ethernal (approved):

vcastellm (approved):

PR #79, opened 2024-09-17
test: Custom native gas token transfer E2E test

Stefan-Ethernal:
## Description

This PR introduces custom native gas token transfer for CDK chain e2e test. The custom native gas token on CDK is originating from the L1 network and is mapped to some existing L1 ERC20 token. Test checks whether native token balance of receiver account increases by the deposited amount on the L2.

The use case is described [here](https://github.com/0xPolygon/kurtosis-cdk/blob/main/docs/native-token/native-token.md).

### DIFF
--- a/test/bridge-e2e.bats
+++ b/test/bridge-e2e.bats
@@ -15,7 +15,7 @@ setup() {
     # Download the genesis file
     readonly bridge_default_address=$(jq -r ".genesis[] | select(.contractName == \"PolygonZkEVMBridge proxy\") | .address" ./tmp/cdk/genesis/genesis.json)
 
-    readonly skey=${SENDER_PRIVATE_KEY:-"12d7de8621a77640c9241b2595ba78ce443d05e94090365ab3bb5e19df82c625"}
+    readonly sender_private_key=${SENDER_PRIVATE_KEY:-"12d7de8621a77640c9241b2595ba78ce443d05e94090365ab3bb5e19df82c625"}
     readonly destination_net=${DESTINATION_NET:-"1"}
     readonly destination_addr=${DESTINATION_ADDRESS:-"0x0bb7AA0b4FdC2D2862c088424260e99ed6299148"}
     readonly ether_value=${ETHER_VALUE:-"0.0200000054"}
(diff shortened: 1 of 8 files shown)


test/helpers/common.bash
```diff
@@ -275,3 +279,54 @@ function checkBalances() {
...
+    local amount_wei=$(cast --to-wei "$ether_amount")
+    
+    # Get final balance in wei (after the operation)
+    local final_balance_wei=$(cast balance "$account" --rpc-url "$rpc_url" | awk '{print $1}')
+
+    # Calculate expected final balance (initial_balance + amount)
+    local expected_final_balance_wei=$(echo "$initial_balance_wei + $amount_wei" | bc)
+
+    # Check if final_balance matches the expected final balance
+    if [ "$(echo "$final_balance_wei == $expected_final_balance_wei" | bc)" -eq 1 ]; then
```
vcastellm, line 303: Shouldn't this be a bats assert?

test/bridge-e2e.bats
```diff
@@ -1,71 +1,128 @@
...
-        fi
-        sleep 10
-    done
-    
+    claim_frequency="10"
+    run wait_for_claim "$timeout" "$claim_frequency"
+    assert_success
+}
+
+@test "Custom native token transfer" {
```
vcastellm, line 58: This test looks so big, can we split it in smaller tests? I feel it can be better

test/helpers/common.bash
Stefan-Ethernal: IDK how to accomplish that with bats `assert`, since the comparison needs to be done with the basic calculator (`bc`) command. Any specific ideas?

test/bridge-e2e.bats
Stefan-Ethernal: If we split the test into smaller ones, it does not make sense. If we could somehow interlock the tests, that would make sense, but to the best of my knowledge, we cannot do it in the Bats...?

I can get rid of assertions for the sender's initial and final balance for the gas token and just blindly assume that minting done what it was suppose to and that balance was increased (in case it returns the function that sends mint transaction yields zero status code). In that case, it would mean, I remove these two parts of the code:
- https://github.com/0xPolygon/cdk/pull/79/files#diff-a6fb205ec4b124e5f87e51865f82e47c846812f102edb1f484deb4c0cb5cdc71R93-R103
- https://github.com/0xPolygon/cdk/pull/79/files#diff-a6fb205ec4b124e5f87e51865f82e47c846812f102edb1f484deb4c0cb5cdc71R80-R83

Does it sound good enough?

goran-ethernal (approved):

PR #84, opened 2024-09-18
fix: Aggregator: use sequenceBatch maxTimestamp as TimestampLimit 

joanestebanr:
## Description

Update `zkevm-synchronizer-l1` library to v1.0.1 to fix the problem
This library have a configuration change, because support SQLLITE

### IMPORTANT NOTE:
**To fix the issue is mantadory to resync all data from Banana**


ToniRamirezM (approved):

goran-ethernal (approved):

config/config.go
```diff
@@ -179,8 +220,45 @@ func Load(ctx *cli.Context) (*Config, error) {
...
+	for _, forbiddenField := range forbiddenFieldsOnConfig {
+		if forbiddenField.FieldName == fieldName {
+			return &forbiddenField
+		}
+	}
+	return nil
+}
 
-	return cfg, nil
+func getUnexpectedFields(keysOnFile, expectedConfigKeys []string) []string {
```
goran-ethernal, line 247: We could turn this `expectedConfigKeys` into the `map[string]struct{}` and then use that instead of `contains` function to get a bit more out of performances.

config/config.go
```diff
@@ -52,6 +51,46 @@ const (
...
+	deprecatedFieldSyncDB = "Aggregator.Synchronizer.DB is deprecated use Aggregator.Synchronizer.SQLDB instead"
+)
+
+type ForbiddenField struct {
+	FieldName string
+	Reason    string
+}
+
+var (
+	forbiddenFieldsOnConfig = []ForbiddenField{
```
goran-ethernal, line 64: I would turn this into a `map[string]ForbidenField` for later easier use when checking if a field is forbidden.

PR #85, opened 2024-09-19
test: Add smart contract e2e tests

rachit77:
## Description

Add smart contract e2e tests using bats


vcastellm (commented):
LGTM but all the assert_output looks too verbose, maybe if one works all work? then we can limit the assertion to just a few meesages?

vcastellm (approved):

Stefan-Ethernal:
FYI: the e2e test is failing due to the issue on the Kurtosis CDK. This is the PR that fixes it (https://github.com/0xPolygon/kurtosis-cdk/pull/276).

test/helpers/common.bash
```diff
@@ -151,21 +151,14 @@ function send_smart_contract_transaction() {
...
-    local sender="$4"
-    shift 4
+    shift 3
     local params=("$@")
 
-    # Verify if the function signature starts with "function"
-    if [[ ! "$function_sig" =~ ^function\ .+\(.+\)$ ]]; then
-        echo "Error: Invalid function signature format '$function_sig'."
-        return 1
-    fi
```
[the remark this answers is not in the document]
rachit77: Don't have strong reason here but I see the more standard way of providing solidity function signature is without the `function` keyword actually

test/basic-e2e.bats
```diff
@@ -8,52 +8,176 @@ setup() {
...
+    # nonce would not increase since transaction fails at the node's pre-validation check
+    # Get bytecode from the contract artifact
+    local bytecode=$(jq -r .bytecode "$contract_artifact")
+    if [[ -z "$bytecode" || "$bytecode" == "null" ]]; then
+        echo "Error: Failed to read bytecode from $contract_artifact"
+        return 1
+    fi
+
+    # Estimate gas, gas price and gas cost
+    local gas_units=$(cast estimate --rpc-url "$l2_rpc_url" --create "$bytecode")
```
Stefan-Ethernal, line 62: WDYT? Btw, with this approach are we going to be able to deduce where exactly the error is if we just return 1? Probably it would be slightly better if we were returning some meaningful error message as well. The same applies to other commands in this scope (especially `cast gas-price`, `cast nonce` etc.)
```suggestion
    local gas_units=$(cast estimate --rpc-url "$l2_rpc_url" --create "$bytecode") || return 1
```

test/basic-e2e.bats
```diff
@@ -8,52 +8,176 @@ setup() {
...
+    gas_units=$(echo "scale=0; $gas_units / 2" | bc)
+    local gas_price=$(cast gas-price --rpc-url "$l2_rpc_url")
+    local value=$(echo "$gas_units * $gas_price" | bc)
+    local value_ether=$(cast to-unit "$value" ether)"ether" 
+
+    # Transfer insufficient funds
```
Stefan-Ethernal, line 68:
```suggestion
    # Transfer only half amount of tokens needed for contract deployment fees
```

test/basic-e2e.bats
```diff
@@ -8,52 +8,176 @@ setup() {
...
+    cast_output=$(cast send --rpc-url "$l2_rpc_url" --private-key "$sender_private_key" "$address_A" --value "$value_ether" --legacy 2>&1)
+    if [[ $? -ne 0 ]]; then
+        echo "Error: Failed to send transaction. Output:"
+        echo "$cast_output"
+        return 1
+    fi
```
Stefan-Ethernal, line 74: Why don't you simply use the `send_tx` function? Seems like we could spare a couple of lines of code. WDYT?

test/basic-e2e.bats
```diff
@@ -8,52 +8,176 @@ setup() {
...
+    ## Case 3: Transaction should fail as address_A tries to transfer more tokens than it has
+    # nonce would not increase 
+    # Transfer funds for gas fees to address_A
+    value_ether="4ether" 
+    cast_output=$(cast send --rpc-url "$l2_rpc_url" --private-key "$sender_private_key" "$address_A" --value "$value_ether" --legacy 2>&1)
+    if [[ $? -ne 0 ]]; then
+        echo "Error: Failed to send transaction. Output:"
+        echo "$cast_output"
+        return 1
+    fi
```
Stefan-Ethernal, line 91: Why don't you simply use the `send_tx` function? Seems like we could spare a couple of lines of code. WDYT?

Stefan-Ethernal (approved):

test/basic-e2e.bats
```diff
@@ -8,52 +8,176 @@ setup() {
...
+
+    run polycli loadtest uniswapv3 --legacy -v 600 --rpc-url $l2_rpc_url --private-key $address_A_private_key
+    assert_success
+
+    # Remove ANSI escape codes from the output
+    output=$(echo "$output" | sed -r "s/\x1B\[[0-9;]*[mGKH]//g")
+
+    # Check if the WETH9 contract were deployed
```
Stefan-Ethernal, line 151:
```suggestion
    # Check if all required Uniswap contracts were deployed
```

test/basic-e2e.bats
rachit77: In the `send_tx` function, there are a few redundant balance checks for the sender and receiver. These checks have already been verified in other test cases, so I decided it would be more efficient to skip them for subsequent transactions to avoid unnecessary repetition.

WDYT?

Stefan-Ethernal: I don't like too much there are the balance checks in the `send_tx`, it should be extracted as a separate function IMO, and invoked only when needed (basically keeping these helper functions focused), but that's probably a separate topic...

For now I'm ok with this, but eventually I would like to never call the cast send explicitly, but rely on the helper function (because it has validations and error checks).

rachit77: Merging this for now, will refactor the part you suggested when we increase the scope e2e tests.

PR #86, opened 2024-09-19
feat: seq sender sanity check l1infotree

joanestebanr:
## Description

When a sequence is build it checks that the L1InfoTreeCount set include the indexes inside batchL2data of all batches conforming the sequence

### DIFF
--- a/sequencesender/sequencesender.go
+++ b/sequencesender/sequencesender.go
@@ -1168,7 +1168,8 @@ func (s *SequenceSender) addInfoSequenceBatchEnd(batch *datastream.BatchEnd) {
 // addNewBatchL2Block adds a new L2 block to the work in progress batch
 func (s *SequenceSender) addNewBatchL2Block(l2Block *datastream.L2Block) {
 	s.mutexSequence.Lock()
-	s.logger.Infof(".....new L2 block, number %d (batch %d)", l2Block.Number, l2Block.BatchNumber)
+	s.logger.Infof(".....new L2 block, number %d (batch %d) l1infotree %d",
+		l2Block.Number, l2Block.BatchNumber, l2Block.L1InfotreeIndex)
 
 	// Current batch
 	data := s.sequenceData[s.wipBatch]
@@ -1183,7 +1184,11 @@ func (s *SequenceSender) addNewBatchL2Block(l2Block *datastream.L2Block) {
 			)
 		}
 		data.batch.SetLastCoinbase(common.BytesToAddress(l2Block.Coinbase))
-		data.batch.SetL1InfoTreeIndex(l2Block.L1InfotreeIndex)
+		if l2Block.L1InfotreeIndex != 0 {
+			data.batch.SetL1InfoTreeIndex(l2Block.L1InfotreeIndex)
+		} else {
+			s.logger.Warnf("L1InfotreeIndex is 0, we don't change batch L1InfotreeIndex (%d)", data.batch.L1InfoTreeIndex())
+		}
 		// New L2 block raw
 		newBlockRaw := state.L2BlockRaw{}
 
--- a/sequencesender/sequencesender.go
+++ b/sequencesender/sequencesender.go
@@ -214,6 +214,7 @@ func (s *SequenceSender) purgeSequences() {
 
 	// Purge the information of batches that are already virtualized
 	s.mutexSequence.Lock()
+	defer s.mutexSequence.Unlock()
 	truncateUntil := 0
 	toPurge := make([]uint64, 0)
 	for i := 0; i < len(s.sequenceList); i++ {
@@ -240,7 +241,6 @@ func (s *SequenceSender) purgeSequences() {
 		}
 		s.logger.Infof("batches purged count: %d, fromBatch: %d, toBatch: %d", len(toPurge), firstPurged, lastPurged)
 	}
-	s.mutexSequence.Unlock()
 }
 
 // purgeEthTx purges transactions from memory structures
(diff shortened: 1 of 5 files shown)


sequencesender/sequencesender.go
```diff
@@ -1168,7 +1168,8 @@ func (s *SequenceSender) addInfoSequenceBatchEnd(batch *datastream.BatchEnd) {
 // addNewBatchL2Block adds a new L2 block to the work in progress batch
 func (s *SequenceSender) addNewBatchL2Block(l2Block *datastream.L2Block) {
 	s.mutexSequence.Lock()
```
Stefan-Ethernal, line 1170: I'm aware that it is unrelated to this PR, but I'm afraid that we are not releasing the lock in case we are returning from the function (https://github.com/0xPolygon/cdk/pull/86/files#diff-9420435b2f8e7ca901e1efd3c5f0dd6c7e267d8a90ae837c135fbd8f239ea5bfR1199-R1202) or even in case we are logging on fatal level (https://github.com/0xPolygon/cdk/pull/86/files#diff-9420435b2f8e7ca901e1efd3c5f0dd6c7e267d8a90ae837c135fbd8f239ea5bfR1181-R1184). So if you don't mind, probably it would be a better idea to unlock on `defer` statement instead of at the end of the function.

```suggestion
	s.mutexSequence.Lock()
	defer s.mutex.Sequence.Unlock()
```

sequencesender/txbuilder/banana_types.go
```diff
@@ -147,3 +148,34 @@ func (b *BananaSequence) LastVirtualBatchNumber() uint64 {
...
+	for i := range batchRawV2.Blocks {
+		if batchRawV2.Blocks[i].IndexL1InfoTree > maxIndex {
+			maxIndex = batchRawV2.Blocks[i].IndexL1InfoTree
+		}
+	}
+	return maxIndex, nil
+}
+
+func CalculateMaxL1InfoTreeIndexInsideSequence(seq *etherman.SequenceBanana) (uint32, error) {
+	maxIndex := uint32(0)
```
Stefan-Ethernal, line 170: Nitpick: consider to validate the `seq` parameter against nil (especially this function is the public one)?

sequencesender/txbuilder/banana_types.go
```diff
@@ -147,3 +148,34 @@ func (b *BananaSequence) LastVirtualBatchNumber() uint64 {
 func (b *BananaSequence) SetLastVirtualBatchNumber(batchNumber uint64) {
 	b.SequenceBanana.LastVirtualBatchNumber = batchNumber
 }
+
+func CalculateMaxL1InfoTreeIndexInsideL2Data(l2data []byte) (uint32, error) {
```
Stefan-Ethernal, line 152: Should we make the function internal?

sequencesender/txbuilder/banana_types.go
```diff
@@ -147,3 +148,34 @@ func (b *BananaSequence) LastVirtualBatchNumber() uint64 {
...
+		return 0, fmt.Errorf("CalculateMaxL1InfoTreeIndexInsideL2Data: error decoding batchL2Data, err:%w", err)
+	}
+	if batchRawV2 == nil {
+		return 0, fmt.Errorf("CalculateMaxL1InfoTreeIndexInsideL2Data: batchRawV2 is nil")
+	}
+	maxIndex := uint32(0)
+	for i := range batchRawV2.Blocks {
+		if batchRawV2.Blocks[i].IndexL1InfoTree > maxIndex {
+			maxIndex = batchRawV2.Blocks[i].IndexL1InfoTree
+		}
```
Stefan-Ethernal, line 164: Minor optimization: use the block loop variable instead of indexing the slice twice. WDYT?

```suggestion
	for _, block := range batchRawV2.Blocks {
		if block.IndexL1InfoTree > maxIndex {
			maxIndex = block.IndexL1InfoTree
```

sequencesender/sequencesender.go
```diff
@@ -1183,7 +1184,11 @@ func (s *SequenceSender) addNewBatchL2Block(l2Block *datastream.L2Block) {
 			)
 		}
 		data.batch.SetLastCoinbase(common.BytesToAddress(l2Block.Coinbase))
-		data.batch.SetL1InfoTreeIndex(l2Block.L1InfotreeIndex)
+		if l2Block.L1InfotreeIndex != 0 {
+			data.batch.SetL1InfoTreeIndex(l2Block.L1InfotreeIndex)
+		} else {
+			s.logger.Warnf("L1InfotreeIndex is 0, we don't change batch L1InfotreeIndex (%d)", data.batch.L1InfoTreeIndex())
```
Stefan-Ethernal, line 1190: Optional:

```suggestion
			s.logger.Warnf("L2 block L1InfoTreeIndex is 0, we don't change batch L1InfoTreeIndex (%d)", data.batch.L1InfoTreeIndex())
```

sequencesender/sequencesender.go
```diff
@@ -1183,7 +1184,11 @@ func (s *SequenceSender) addNewBatchL2Block(l2Block *datastream.L2Block) {
 			)
 		}
 		data.batch.SetLastCoinbase(common.BytesToAddress(l2Block.Coinbase))
-		data.batch.SetL1InfoTreeIndex(l2Block.L1InfotreeIndex)
+		if l2Block.L1InfotreeIndex != 0 {
+			data.batch.SetL1InfoTreeIndex(l2Block.L1InfotreeIndex)
```
Stefan-Ethernal, line 1188: Is it expected to overwrite the `L1InfotreeIndex` in case it is already set for the batch?

Stefan-Ethernal (approved):
Generally LGTM, left a couple of comments to consider.

sequencesender/sequencesender.go
joanestebanr: good catch, I have also changed any unlock with a defer one

sequencesender/sequencesender.go
joanestebanr: I think so, is the previous implementation. I just add that is the incomming L2Block have 0 don't change it in the batch

goran-ethernal (approved):

vcastellm (approved):
LGTM

PR #95, opened 2024-09-26
feat: Sequence sender unit tests

rbpol:


.mockery.yaml
```diff
@@ -0,0 +1,25 @@
+with-expecter: true
+filename: "{{.InterfaceName | lower }}.generated.go"
```
rachit77, line 2:
```suggestion
filename: "{{.InterfaceName | lower }}_mock_test.go"
```

How about just defining the file name here globally?

.mockery.yaml
```diff
@@ -0,0 +1,25 @@
...
+mockname: "{{.InterfaceName}}Mock"
+packages:
+  github.com/0xPolygon/cdk/sequencesender:
+    config:
+      dir: ./sequencesender
```
rachit77, line 7: should we place mocks in a separate dir.
Maybe `./sequencesender/mocks`

WDYT?

rbpol: Cannot do that due to a circle dependencies

.mockery.yaml
rbpol: Cannot do that if we move mocks to "mocks" dir. See below.

.mockery.yaml
rbpol: Done

Makefile
```diff
@@ -83,6 +90,13 @@ test-seq_sender:
 lint: ## Runs the linter
 	export "GOROOT=$$(go env GOROOT)" && $$(go env GOPATH)/bin/golangci-lint run --timeout 5m
 
+.PHONY: generate
+generate: generate-code-from-proto generate-mocks ## Generates mocks and other autogenerated types
+
+.PHONY: generate-mocks
+generate-mocks: ## Generates mocks
```
joanestebanr, line 97: The target `generate-mocks`  is  inside `test/Makefile`. Please, check it. By other hand the naming of the generated mocks doesn't match the rest

rbpol: Fixed

rbpol:
https://github.com/0xPolygon/cdk/pull/103

PR #99, opened 2024-10-01
feat: use config-file on CDK for kurtosis e2e tests

joanestebanr:
## Description

- The e2e test for CDK use a local configuration template instead of the one on kurtosis branch: to allow to develop changes on config file is more easy if we are able to use the config file on CDK repo. After that we can update config file and cdk image on kurtosis repo
- The local configuration for run on vscode use the same template (`test/config/kurtosis-cdk-node-config.toml.template`)  as the e2e test
- Fix error on CI `test-e2e.yml` with the duplicated `ref` key


vcastellm (approved):
LGTM

Stefan-Ethernal (approved):

PR #101, opened 2024-10-01
refactor: simplify running Kurtosis combinations

vcastellm:
Using a set of pre-defined combinations of components, we're going to test the versions we're interested in.

* Standardize everything to fork ids instead of names, this is to match Kurtosis.
* Also start testing Fork12


test/scripts/kurtosis_prepare_params_yml.sh
```diff
@@ -1,31 +0,0 @@
...
-if [ -z $DATA_AVAILABILITY_MODE ]; then
-    echo "Missing param Data Availability Mode : [rollup, cdk-validium]"
-    exit 1
-fi
-
-mkdir -p $(dirname $DEST_KURTOSIS_PARAMS_YML)
-cp $KURTOSIS_FOLDER/params.yml $DEST_KURTOSIS_PARAMS_YML
-yq -Y --in-place ".args.cdk_node_image = \"cdk\"" $DEST_KURTOSIS_PARAMS_YML
-yq -Y --in-place ".args.data_availability_mode = \"$DATA_AVAILABILITY_MODE\"" $DEST_KURTOSIS_PARAMS_YML
-yq -Y --in-place ".args.zkevm_use_gas_token_contract = $USE_L1_GAS_TOKEN_CONTRACT" $DEST_KURTOSIS_PARAMS_YML
```
Stefan-Ethernal, line 31: I certainly need this to tweak the configuration and use the L1 gas token which is necessary for the custom L1 gas token E2E test.

vcastellm: Let's prepare a new template?

vcastellm: I ended up adding the config param to the different combinations files

scripts/local_config
```diff
@@ -178,7 +178,7 @@ function check_requirements(){
     cat << EOF
             echo "deb [trusted=yes] https://apt.fury.io/kurtosis-tech/ /" | sudo tee /etc/apt/sources.list.d/kurtosis.list
         echo "deb [trusted=yes] https://apt.fury.io/kurtosis-tech/ /" | sudo tee /etc/apt/sources.list.d/kurtosis.list
-        sudo apt install kurtosis-cli=0.90.1
```
joanestebanr, line 181: `test-e2e.yml` is installing `kurtosis-cli=1.3.0`... doesn't make sense to set the same version here?

test/Makefile
```diff
@@ -46,14 +46,24 @@ generate-mocks-sync: ## Generates mocks for sync, using mockery tool
...
-.PHONY: test-e2e-elderberry-rollup
-test-e2e-elderberry-rollup: stop ## Runs e2e tests checking elderberry/rollup
-	./run-e2e.sh rollup
+.PHONY: test-e2e-fork9-rollup
+test-e2e-fork9-rollup: stop ## Runs e2e tests checking elderberry/rollup
+	./run-e2e.sh fork9 rollup
+	bats .
+
+.PHONY: test-e2e-fork12-validium
+test-e2e-fork12-validium: stop ## Runs e2e tests checking elderberry/validium
```
joanestebanr, line 60: The comment is wrong: must be banana/validium

test/Makefile
```diff
@@ -46,14 +46,24 @@ generate-mocks-sync: ## Generates mocks for sync, using mockery tool
...
+	./run-e2e.sh fork12 cdk-validium
+	bats .
+
+.PHONY: test-e2e-fork12-rollup
+test-e2e-fork12-rollup: stop ## Runs e2e tests checking elderberry/rollup
```
joanestebanr, line 65: The comment is wrong: must be banana/rollup

test/combinations/fork12-cdk-validium.yml
```diff
@@ -0,0 +1,9 @@
+args:
```
joanestebanr, line 1: how do you set forkid = 12?

scripts/local_config
vcastellm: I think it's better to use the latest version in local so it will force it to stay up to date. 

Notice that when you run in the terminal it will output a warning on the outdated local installation and the suggestion to upgrade, so should do it anyway, and in case there's an incompatibility in the CI you should manually upgrade the CI too.

test/combinations/fork12-cdk-validium.yml
vcastellm: The fork is set in the deployed contracts image, and in input_parser.star in kurtosis there's the logic that sets the fork in the config.

test/Makefile
vcastellm: Removed the redundant comments

joanestebanr (approved):

.github/workflows/test-e2e.yml
```diff
@@ -35,7 +39,7 @@ jobs:
       run: |
         echo "deb [trusted=yes] https://apt.fury.io/kurtosis-tech/ /" | sudo tee /etc/apt/sources.list.d/kurtosis.list
         sudo apt update
-        sudo apt install kurtosis-cli=0.90.1
+        sudo apt install kurtosis-cli=1.3.0
```
Stefan-Ethernal, line 42: Should we just use the latest maybe?

Stefan-Ethernal (approved):
Generally LGTM.

.github/workflows/test-e2e.yml
vcastellm: In the case of the CI we want deterministic builds, you know... this is why my response to @joanestebanr, it should be updated by us manually

PR #104, opened 2024-10-03
feat: adapt kurtosis cdk and fix errors

joanestebanr:
## Description

Last changes on `kurtosis-cdk` main brach have changed the way that kurtosis answer to ports and also have change some end-points names. 
- Some request of ports return the protocol and another no so the code have been updated to support port with protocol or without (`http://127.0.0.1:8023` or `127.0.0.1:8023`) 
- Also, to support changing end-point names have include a list of 'end-points' to get the URL / port
- There was a hardcoded path in config file that have been changed by vars and left the absolute path used by kurtosis as deault value

### DIFF
--- a/scripts/local_config
+++ b/scripts/local_config
@@ -176,8 +176,8 @@ function export_value_from_kurtosis_or_fail(){
     done
     export $_EXPORTED_VAR_NAME=$_RESULT
     if [ -z $_EXPORTED_VAR_NAME ]; then
-            log_fatal "Error getting kurtosis port:  $KURTOSIS_ENCLAVE $_SERVICE $_END_POINT"
-        fi
+        log_fatal "Error getting kurtosis port:  $KURTOSIS_ENCLAVE $_SERVICE $_END_POINT"
+    fi
     log_debug "$_EXPORTED_VAR_NAME=${!_EXPORTED_VAR_NAME}  \t\t\t# Kurtosis $KURTOSIS_ENCLAVE $_SERVICE $_END_POINT"
 }
 ###############################################################################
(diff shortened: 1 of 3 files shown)


vcastellm (approved):
LGTM

scripts/local_config
```diff
@@ -37,12 +38,23 @@ function get_value_from_toml_file(){
             if [[ "$_LINE" == [* ]]; then
                 return 1
             fi
+            if [ $_return_next_line -eq 1 ]; then
+               # sed sentence remove quotes
+               echo $_LINE |  sed 's/^[[:space:]]*"//;s/"$//'
```
Stefan-Ethernal, line 43: An extra space, not needed?
```suggestion
               echo $_LINE | sed 's/^[[:space:]]*"//;s/"$//'
```

scripts/local_config
```diff
@@ -126,25 +149,46 @@ function export_values_of_cdk_node_config(){
...
+        if [ ! -z $_RESULT ]; then
+            break
+        fi
+    done
+    export $_EXPORTED_VAR_NAME=$_RESULT
     if [ -z $_EXPORTED_VAR_NAME ]; then
-        log_fatal "Error getting kurtosis port:  $KURTOSIS_ENCLAVE $_SERVICE $_END_POINT"
-    fi
+            log_fatal "Error getting kurtosis port:  $KURTOSIS_ENCLAVE $_SERVICE $_END_POINT"
+        fi
```
Stefan-Ethernal, line 180: Is there an extra tab in both lines (misaligned with the `if`)?

```suggestion
         log_fatal "Error getting kurtosis port:  $KURTOSIS_ENCLAVE $_SERVICE $_END_POINT"
     fi
```

Stefan-Ethernal (approved):

PR #105, opened 2024-10-03
feat: Proof Cleaning

ToniRamirezM:
## Description

Cleans generated proofs once they have been sent to L1 and its tx is finalized.


aggregator/aggregator.go
```diff
@@ -1928,57 +1929,39 @@ func (hc *healthChecker) Watch(req *grpchealth.HealthCheckRequest, server grpche
...
-	   // network is synced with the final proof, we can safely delete all recursive
-	   // proofs up to the last synced batch
-	   err = a.State.CleanupGeneratedProofs(a.ctx, proofBatchNumberFinal, nil)
+	for result.MinedAtBlockNumber.Uint64() > finaLizedBlockNumber {
+		select {
+		case <-a.ctx.Done():
+			return
+		case <-time.After(a.cfg.RetryTime.Duration):
+			finaLizedBlockNumber, err = l1_check_block.L1FinalizedFetch.BlockNumber(a.ctx, a.etherman)
+			if err != nil {
```
joanestebanr, line 1952: missing a return for this err?

ToniRamirezM: Te functions does not return error. The idea is if it fails here, do another iteration of the loop.

Stefan-Ethernal (approved):

joanestebanr (approved):

PR #111, opened 2024-10-08
feat: make the fetching of extra claim data optional for bridge sync

arnaubennassar:
Make fetching the full claim data optional. This is needed since:
- it breaks on pre-etrog chains
- should be used only for PP chains


Stefan-Ethernal (approved):

goran-ethernal (approved):

arnaubennassar:
deprecated

PR #114, opened 2024-10-08
ci: fix release tags

vcastellm:
Allow release from flexible tags

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,9 +33,6 @@ jobs:
       TAGS: ${{ steps.meta.outputs.tags }}
       VERSION: ${{ steps.meta.outputs.version }}
     steps:
-      - name: validate tag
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,10 +1,10 @@
 name: Release
 
 on:
-    push:
-        tags:
-        # run only against tags that follow semver (https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string)
-        - 'v[0-9]+.[0-9]+.[0-9]+*'
+  push:
+    tags:
+      - 'v*.*.*'
```
Stefan-Ethernal, line 6: IMO it is better to keep the previous regex, as it is more common way to name versions and star allows to have whatever suffix we want. WDYT?

Stefan-Ethernal (approved):

rachit77 (approved):

PR #120, opened 2024-10-09
feat: add versions command

vcastellm:
This PR splits the previously global required config and genesis flags to different subcommands as not all commands need the same params.

`node` command is getting the node config file
`erigon` command is getting the config file and the genesis file
`versions` command doesn't need any param

Introducing here the `versions` command, simply outputs the corresponding versions of other components. This is a temporary launch requirement until we iterate on more comprehensive status checks.

### DIFF
--- a/crates/cdk/src/versions.rs
+++ b/crates/cdk/src/versions.rs
@@ -1,32 +1,47 @@
-/// Command to print corresponding versions to the stdout.
+use colored::*;
+use execute::Execute;
+use std::io;
+use std::process::{Command, Output};
+
+fn version() -> Result<Output, io::Error> {
+    let bin_path = crate::helpers::get_bin_path();
+
+    // Run the node passing the config file path as argument
+    let mut command = Command::new(bin_path.clone());
+    command.args(&["version"]);
+
+    command.execute_output()
+}
+
 pub(crate) fn versions() {
-    // Multi-line string to print the versions.
-    println!(
-        r#"
-# FORK 12 IMAGES
-# yq .args .github/tests/forks/fork12.yml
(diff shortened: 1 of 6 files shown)


vcastellm:
Currently it's a mixed format, we should decide which one:

image

zuiris (approved):

crates/cdk/src/versions.rs
```diff
@@ -0,0 +1,47 @@
...
+        ("zkEVM Prover", "v8.0.0-RC12"),
+        ("CDK Erigon", "hermeznetwork/cdk-erigon:0948e33"),
+        (
+            "zkEVM Pool Manager",
+            "hermeznetwork/zkevm-pool-manager:v0.1.1",
+        ),
+        (
+            "CDK Data Availability Node",
+            "0xpolygon/cdk-data-availability:0.0.10",
+        ),
```
Stefan-Ethernal, line 37: How do we ensure that we don't forget to update these? My concern is that soon this will be outdated and we are going to forget to update the specific version, because for example we have affected the `zkevm-contract` or `cdk-erigon`, so we need to raise the PR on the cdk repo (which is probably be forgotten).

IMO it would make more sense if each component in the CDK stack exposes the `version` command and then if possible having a command/script in the Kurtosis CDK repo, that will print the versions of all the components.

It makes sense that the `version` command in the CDK prints only the CDK node version.

WDYT?

vcastellm: I've the same concerns but we've been requested to hadcode this at this point. Not my call

Stefan-Ethernal (approved):

PR #138, opened 2024-10-22
feat: remove DS from aggregator

ToniRamirezM:
## Description

Removes DS from Aggregator.


rbpol (approved):

rpc/types/rpcbatch.go
```diff
@@ -9,34 +9,44 @@ import (
...
+	accInputHash         common.Hash    `json:"accInputHash"`
 	blockHashes          []string       `json:"blocks"`
 	batchL2Data          []byte         `json:"batchL2Data"`
 	globalExitRoot       common.Hash    `json:"globalExitRoot"`
+	localExitRoot        common.Hash    `json:"localExitRoot"`
+	stateRoot            common.Hash    `json:"stateRoot"`
 	coinbase             common.Address `json:"coinbase"`
 	closed               bool           `json:"closed"`
 	lastL2BlockTimestamp uint64         `json:"lastL2BlockTimestamp"`
 	l1InfoTreeIndex      uint32         `json:"l1InfoTreeIndex"`
```
Stefan-Ethernal, line 21: A bit unrelated to this PR, but should we export these fields? Warning analysis is complaining.

rpc/batch.go
```diff
@@ -93,3 +109,39 @@ func getL2BlockTimestampFromRPC(addr, blockHash string) (uint64, error) {
...
+	log.Infof("Requesting witness for batch %d of type %s", batchNumber, witnessType)
+
+	response, err = rpc.JSONRPCCall(b.url, "zkevm_getBatchWitness", batchNumber, witnessType)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check if the response is an error
+	if response.Error != nil {
+		if response.Error.Message == "busy" {
```
Stefan-Ethernal, line 134: Nitpick: extract the `busy` string to a var?

aggregator/aggregator.go
```diff
@@ -289,33 +219,13 @@ func (a *Aggregator) handleReorg(reorgData synchronizer.ReorgExecutionResult) {
...
-
 	// Get new last verified batch number from L1
 	var lastVerifiedBatchNumber uint64
-	if err == nil {
-		lastVerifiedBatchNumber, err = a.etherman.GetLatestVerifiedBatchNum()
-		if err != nil {
-			a.logger.Errorf("Error getting latest verified batch number: %v", err)
-		}
+	lastVerifiedBatchNumber, err = a.etherman.GetLatestVerifiedBatchNum()
+	if err != nil {
```
Stefan-Ethernal, line 227: Nitpick, simplification:

```suggestion

	// Get new last verified batch number from L1
	lastVerifiedBatchNumber, err := a.etherman.GetLatestVerifiedBatchNum()
```

Stefan-Ethernal (commented):
Generally LGTM, leaving a couple of comments to consider.

Stefan-Ethernal (approved):

crates/cdk/src/config_render.rs
```diff
@@ -93,7 +93,6 @@ fn render_yaml(config: &Config, res: Rendered) -> String {
 chain: dynamic-{chain_id}
 zkevm.l2-chain-id: {chain_id}
 zkevm.l2-sequencer-rpc-url: {l2_sequencer_rpc_url}
-zkevm.l2-datastreamer-url: {datastreamer_host}
 zkevm.l1-chain-id: {l1_chain_id}
```
vcastellm, line 96: this is still needed for cdk-erigon node, but we removed the last ocurrence in the node config so IDK from where we're going to get this 😅

config/default.go
```diff
@@ -133,13 +132,12 @@ SenderAddress = "{{SenderProofToL1Addr}}"
...
 GasOffset = 0
+RPCURL = "{{L2URL}}"
 WitnessURL = "{{WitnessURL}}"
-UseL1BatchData = true
 UseFullWitness = false
 SettlementBackend = "l1"
 AggLayerTxTimeout = "5m"
 AggLayerURL = "{{AggLayerURL}}"
-MaxWitnessRetrievalWorkers = 2
 SyncModeOnlyEnabled = false
```
vcastellm, line 141: aren't we using this anymore?

crates/cdk/src/config_render.rs
vcastellm: self-note: We'll guess it from RPC url with a default value

vcastellm (approved):
I'm just missing removing the datastremer lib from go.mod, is that possible now?

config/default.go
ToniRamirezM: Nop, now the witness is requestes on the same thread a prover opens when connecting to the aggregator. CDK-Erigon is protected in case we request to much witnesses with the "busy" logic.

aggregator/config.go
```diff
@@ -112,21 +112,18 @@ type Config struct {
 	// final gas: 1100
 	GasOffset uint64 `mapstructure:"GasOffset"`
 
+	// RPCURL is the URL of the RPC server
+	RPCURL string `mapstructure:"RPCURL"`
```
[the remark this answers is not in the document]
ToniRamirezM: I tested locally on Kurtosis and it looks ok to me.

ToniRamirezM:
> I'm just missing removing the datastremer lib from go.mod, is that possible now?
> 

Very good point 😅 Just removed it now along and unused file in the state package that was importing it.

PR #144, opened 2024-10-25
feat: epoch notifier

joanestebanr:
## Description
`AggLayer` require 1 certificate per network each epoch.  So the idea is before closing the current AggLayer epoch send the certificate. To do that we add a new object `
- Send certificates near end of AggLayer Epoch:
    - Integrate auto-config feature using AggLayer call: `interop_getClockConfiguration`
    - Receive the new block from another object called: `BlockNotifier`

## Internals
- Mocks files are generated automatically for all interfaces of `aggsender` without `_mock` suffix on filename
- 

## Configuration changes:
None of this new fields are mandatory, you can use default ones.
```
[AggSender]
(...)
BlockFinality = "LatestBlock"
EpochNotificationPercentage = 50
```
- `BlockFinality`: is the point that we check the block to decice the epoch. Currently `AggLayer` is using **latest**
- `EpochNotificationPercentage`: It generate a new certificate `BlocksBeforeEpochEnding` blocks after reaching this percentage of epoch. Percent is 0 begining, 100 end

## Tests
- This version have been checked using AggLayer: `ghcr.io/agglayer/agglayer:0.2.0-rc.9`


cmd/run.go
```diff
@@ -144,13 +145,31 @@ func start(cliCtx *cli.Context) error {
...
+		BlockFinalityType:     etherman.LatestBlock,
+		CheckNewBlockInterval: aggsender.AutomaticBlockInterval,
+	}, logger, nil)
+	if err != nil {
+		return nil, err
+	}
+	epochNotifier, err := aggsender.NewEpochNotifierPerBlock(
+		blockNotifier,
+		logger,
+		aggsender.ConfigEpochNotifierPerBlock{
```
goran-ethernal, line 163: I think we should move the two configs for notifiers to the real config, and make it a part of `Aggsender` config part in `.toml` files, like:

```
[Aggsender.EpochNotifier]
BlockFinality = "Latest"
CheckNewBlockInterval = "1s"
StartingEpochBlock = 1000
EpochSize = 10
BlocksBeforeEpochEnding = 2
```

cmd/run.go
```diff
@@ -144,13 +145,31 @@ func start(cliCtx *cli.Context) error {
...
+			StartingEpochBlock:                0,
```
goran-ethernal, line 164: I think we need to set this to a non zero value. I've raised this question with the agglayer team couple of weeks ago, but didn't get the right response. We need to know from which L1 block the agglayer starts counting epochs. If they start let's say at block 1010, and we don't set the `StartingEpochBlock` to this, we might get wrong results when calculating epoch and when to send a certificate.

aggsender/types/generic_subscriber.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+type SubscriptionHandler[T any] struct {
```
goran-ethernal, line 3: Is this used? I can't seem to find any usage of this struct.

aggsender/types/epoch_notifier.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+type EpochEvent struct {
```
goran-ethernal, line 3: If we only need the epoch then I would remove this event, and just return an `uint64` in the channel.

aggsender/types/generic_subscriber.go
joanestebanr: You're right no longer used!

aggsender/types/epoch_notifier.go
joanestebanr: I prefer the struct because we can add more info in a easy way if we need and also is 'hard' type

cmd/run.go
joanestebanr: I think that is better to keep config-file simple: the internal structure it's better to be decoupled with config-file

aggsender/block_notifier_polling.go
```diff
@@ -0,0 +1,181 @@
...
+	if status != nil {
+		res += fmt.Sprintf(" lastBlockSeen=%d", status.lastBlockSeen)
+	} else {
+		res += " lastBlockSeen=none"
+	}
+	return res
+}
+
+func (b *BlockNotifierPolling) Start(ctx context.Context) {
+	go b.start(ctx)
```
arnaubennassar, line 72: In general, the pattern is to let the `cmd` or other "top level" package to decide when to use `go`.

So I'd suggest to remove this function and make public the `start` func

aggsender/epoch_notifier_per_block.go
```diff
@@ -0,0 +1,130 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/0xPolygon/cdk/aggsender/types"
+)
+
+const (
+	// EthereumBlockSpeed is the speed of the Ethereum to generate a blocks
+	EthereumBlockSpeed = time.Second * 15
```
arnaubennassar, line 13:
```suggestion
	EthereumBlockSpeed = time.Second * 12
```

arnaubennassar (changes requested):

aggsender/epoch_notifier_per_block.go
joanestebanr: You're right! but I'll remove this const because is no longer used, the sampling frequency is calculated dynamically

aggsender/block_notifier_polling.go
joanestebanr: Ok, I see  that `AggSender` have the same issue:  launch the go routines inside the `Start` call, In that case it launch 2 go routines. How do you suggest to fix that?

aggsender/block_notifier_polling.go
```diff
@@ -0,0 +1,181 @@
...
+	ethClient     types.EthClient
+	blockFinality *big.Int
+	logger        types.Logger
+	config        ConfigBlockNotifierPolling
+	mu            sync.Mutex
+	lastStatus    *blockNotifierPollingInternalStatus
+	types.GenericSubscriber[types.EventNewBlock]
+}
+
+func NewBlockNotifierPolling(ethClient types.EthClient,
```
vcastellm, line 39: Can you add a description of what this does?

aggsender/block_notifier_polling.go
vcastellm: Could be the same, move the go routines to the top level?

agglayer/client.go
```diff
@@ -128,3 +133,22 @@ func (c *AggLayerClient) GetCertificateHeader(certificateHash common.Hash) (*Cer
 
 	return result, nil
 }
+
+func (c *AggLayerClient) GetClockConfiguration() (*ClockConfiguration, error) {
```
goran-ethernal, line 137: Add a comment to this function, to be consistent with others in this file.

agglayer/types.go
```diff
@@ -385,3 +385,12 @@ func (c CertificateHeader) String() string {
 	return fmt.Sprintf("Height: %d, CertificateID: %s, NewLocalExitRoot: %s",
 		c.Height, c.CertificateID.String(), c.NewLocalExitRoot.String())
 }
+
+type ClockConfiguration struct {
```
goran-ethernal, line 389: Add comments to be consistent with other structs and functions in the file.

cmd/run.go
```diff
@@ -144,13 +145,32 @@ func start(cliCtx *cli.Context) error {
...
+	}, logger, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	notifierCfg, err := aggsender.NewConfigEpochNotifierPerBlock(agglayerClient, cfg.BlocksBeforeEpochEnding)
+	if err != nil {
+		return nil, fmt.Errorf("cant generate config for Epoch Notifier because: %w", err)
+	}
+	epochNotifier, err := aggsender.NewEpochNotifierPerBlock(
```
goran-ethernal, line 165: Should we start the epoch notifier here. I don't see it being started anywhere.

cmd/run.go
```diff
@@ -144,13 +145,32 @@ func start(cliCtx *cli.Context) error {
...
 	cfg aggsender.Config,
+	l1EthClient *ethclient.Client,
 	l1InfoTreeSync *l1infotreesync.L1InfoTreeSync,
-	l2Syncer *bridgesync.BridgeSync,
-) (*aggsender.AggSender, error) {
+	l2Syncer *bridgesync.BridgeSync) (*aggsender.AggSender, error) {
 	logger := log.WithFields("module", cdkcommon.AGGSENDER)
 	agglayerClient := agglayer.NewAggLayerClient(cfg.AggLayerURL)
+	blockNotifier, err := aggsender.NewBlockNotifierPolling(l1EthClient, aggsender.ConfigBlockNotifierPolling{
+		BlockFinalityType:     etherman.LatestBlock,
```
goran-ethernal, line 154: I am unsure about polling the latest block. We might need to check how agglayer does this, and how they count blocks in epoch. Do they poll for latest or finalized blocks.

joanestebanr: maybe we can move that to configuration

goran-ethernal: I've saw their answer. Its Latest block, so we are ok.

goran-ethernal (approved):

vcastellm (approved):
LGTM

aggsender/block_notifier_polling_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/0xPolygon/cdk/etherman"
+	"github.com/0xPolygon/cdk/log"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExploratoryBlockNotifierPolling(t *testing.T) {
+	t.Skip()
```
vcastellm, line 21: should we remove?

joanestebanr: it's a exploratory test, just for test it if we need

test/bridge-e2e.bats
```diff
@@ -155,3 +155,30 @@ setup() {
     fi
     assert_success
 }
+
+@test "transfer ETH L1 to L2 to L1" {
```
vcastellm, line 159: You can merge this with `"Native gas token deposit to WETH"` tests at the top of the file as commented

aggsender/types/block_notifier.go
```diff
@@ -0,0 +1,14 @@
...
+
+import "github.com/0xPolygon/cdk/etherman"
+
+type EventNewBlock struct {
+	BlockNumber       uint64
+	BlockFinalityType etherman.BlockNumberFinality
+}
+
+type BlockNotifier interface {
+	// NotifyEpochStarted notifies the epoch has started.
```
vcastellm, line 11: Does this line belongs to here?

vcastellm (changes requested):
Observe some new comments

aggsender/types/block_notifier.go
joanestebanr: The interface `BlockNotifier` is the interface that use `EpochNotifierPerBlock`. Now there are 1 implementation that is `BlockNotifierPolling` but the idea is to create another one  using WS and decide in runtime which one to use

aggsender/aggsender.go
```diff
@@ -475,19 +478,6 @@ func (a *AggSender) signCertificate(certificate *agglayer.Certificate) (*agglaye
...
-			a.checkPendingCertificatesStatus(ctx)
-		case <-ctx.Done():
-			return
-		}
-	}
-}
-
 // checkPendingCertificatesStatus checks the status of pending certificates
 // and updates in the storage if it changed on agglayer
 func (a *AggSender) checkPendingCertificatesStatus(ctx context.Context) {
```
goran-ethernal, line 483: I think we need to change this function to `anyPendingCertificates` and return a bool and an error. Bool meaning if we have pending certificates, and error, if some error occurs while checking. Then if the bool is true, we just continue the loop in `sendCertificates`, and if error, just log it and continue the loop.

If we leave it as it is now, we can have pending certificates, and we will try to send a new one.

agglayer/client.go
```diff
@@ -15,14 +15,22 @@ import (
 
 const errCodeAgglayerRateLimitExceeded int = -10007
 
-var ErrAgglayerRateLimitExceeded = fmt.Errorf("agglayer rate limit exceeded")
+var (
+	ErrAgglayerRateLimitExceeded = fmt.Errorf("agglayer rate limit exceeded")
+	jSONRPCCall                  = rpc.JSONRPCCall
```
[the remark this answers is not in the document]
joanestebanr: Yes, is for unittest the client

vcastellm (approved):

PR #150, opened 2024-10-30
feat: use sqlite on lastgersync

arnaubennassar:


lastgersync/processor.go
```diff
@@ -2,292 +2,135 @@ package lastgersync
...
-			log.Errorf("unexpected type %T in events", e)
+			return errors.New("failed to convert sync.Block.Event to Event")
 		}
-		if int64(event.L1InfoTreeIndex) < lastIndex {
-			continue
+		type eventWithBlockNum struct {
+			GlobalExitRoot  ethCommon.Hash `meddler:"global_exit_root,hash"`
+			L1InfoTreeIndex uint32         `meddler:"l1_info_tree_index"`
+			BlockNum        uint64         `meddler:"block_num"`
 		}
```
Stefan-Ethernal, line 94: Can we move the struct declaration outside of the loop (or even outside of the `ProcessBlock` function), as it is unnecessary declaring it multiple times.

lastgersync/migrations/lastgersync0001.sql
```diff
@@ -0,0 +1,14 @@
+-- +migrate Down
+DROP TABLE IF EXISTS block;
+DROP TABLE IF EXISTS global_exit_root;
+
+-- +migrate Up
+CREATE TABLE block (
+    num   BIGINT PRIMARY KEY
+);
```
Stefan-Ethernal, line 8: What is the purpose of this table? Do you think we are going to expand it furthermore in the future and therefore it is adjusted to leave it? Otherwise, IMO we should be good enough with having only the `global_exit_root` table. WDYT?

lastgersync/processor.go
```diff
@@ -2,292 +2,135 @@ package lastgersync
...
+			return 0, ethCommon.Hash{}, db.ErrNotFound
+		}
 		return 0, ethCommon.Hash{}, err
 	}
-	if l1InfoIndexBytes == nil {
-		return 0, ethCommon.Hash{}, db.ErrNotFound
-	}
-
-	return common.BytesToUint32(l1InfoIndexBytes), ethCommon.BytesToHash(ger), nil
+	return e.L1InfoTreeIndex, e.GlobalExitRoot, nil
```
Stefan-Ethernal, line 135: Nitpick:
maybe it would be simpler if we return the `Event` instance?

lastgersync/processor.go
```diff
@@ -2,292 +2,135 @@ package lastgersync
...
-			tx.Rollback()
-			return err
-
-		default:
-			lastIndex = int64(li)
+	shouldRollback := true
+	defer func() {
+		if shouldRollback {
+			if errRllbck := tx.Rollback(); errRllbck != nil {
+				log.Errorf("error while rolling back tx %v", errRllbck)
```
Stefan-Ethernal, line 77: Nitpick, only 2 letters longer but nicer IMO, if you agree (not pushing, just a suggestion though). :slightly_smiling_face:
```suggestion
			if errRollback := tx.Rollback(); errRollback != nil {
				log.Errorf("error while rolling back tx %v", errRollback)
```

Stefan-Ethernal (approved):
Generally LGTM!

lastgersync/migrations/lastgersync0001.sql
arnaubennassar: this table is used to handle reorgs and to track the status of syncing: some time there is no update that affect GERs, but yet we have synced this blocks. We could rely only on the GER table, and make the actual info of the table be NULLABLE, and add a bit more logic into the queries.

I like more the block table, as it's consistent with the other syncers. But I'm open to change this, WDYT?

Stefan-Ethernal: In hindsight, it is probably better to leave it, though it is a bit strange to have only a single column (who knows, maybe we will store some more data in it at some point).

goran-ethernal (approved):

vcastellm (approved):
LGTM

rachit77 (approved):

PR #155, opened 2024-10-31
Fix/revert l1infotree changes

joanestebanr:
## Description

Please include a summary of the change and which issue is fixed. Also, include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)


joanestebanr:
The revert have been done by PR: https://github.com/0xPolygon/cdk/pull/174

PR #158, opened 2024-11-01
feat: unpack and log agglayer errors

goran-ethernal:
## Description

This PR introduces error unpacking when getting a certificate header using the `interop_getCertificateHeader` rpc endpoint on `agglayer`. 

When `agglayer` marks a certificate as `InError` it also packs the error in the json format of `CertificateHeader` status field. These errors can be different, where each error has its own unique error code and structure.

This PR handles unmarshaling of these different errors, so we can log them when we get the certificate header from `agglayer`, which can be useful for debugging purposes, to know in a clear way what was the cause of error when settling a certificate.


arnaubennassar (approved):

agglayer/client.go
```diff
@@ -91,6 +91,8 @@ func (c *AggLayerClient) WaitTxToBeMined(hash common.Hash, ctx context.Context)
 
 // SendCertificate sends a certificate to the AggLayer
 func (c *AggLayerClient) SendCertificate(certificate *SignedCertificate) (common.Hash, error) {
+	certificate.ImportedBridgeExits = []*ImportedBridgeExit{}
+	certificate.BridgeExits = []*BridgeExit{}
```
Stefan-Ethernal, line 95: IIRC this is changed in the #165 (`CopyWithDefaults` function)? If that's correct, we can probably revert this in favor of #165 changes.

agglayer/proof_generation_error.go
```diff
@@ -0,0 +1,657 @@
...
+)
+
+type PPError interface {
+	String() string
+}
+
+// ProofGenerationError is a struct that represents an error that occurs when generating a proof.
+type ProofGenerationError struct {
+	GenerationType string
+	InnerErrors    []PPError
```
Stefan-Ethernal, line 44: Do you even need the `PPError` interface?

```suggestion
	InnerErrors    []error
```

Stefan-Ethernal (approved):

agglayer/proof_generation_error.go
goran-ethernal: I need an interface because of different errors (different structs). I need a way to unmarshal different errors, and that is why I need an interface here, since there are 50+ different errors.

PR #161, opened 2024-11-05
refactor: retrieve and parse versions at buildtime

vcastellm:
Use input_parser.star from kurtosis

This is the output

image


crates/cdk/build.rs
```diff
@@ -46,3 +53,58 @@ fn main() {
...
+    let response = get(url).expect("Failed to send request");
+    let content = response.text().expect("Failed to read response text");
+
+    // Write the contents to a file
+    let out_dir = std::env::var("OUT_DIR").unwrap();
+    let dest_path = Path::new(&out_dir).join("input_parser.star");
+    let mut file = File::create(&dest_path)?;
+    file.write_all(content.as_bytes())?;
+
+    // Get lines 28 to 40 from the contents of the starlark file
```
leovct, line 70: This looks very dangerous to me!

crates/cdk/build.rs
```diff
@@ -46,3 +53,58 @@ fn main() {
...
+    let versions = content
+        .lines()
+        .skip(30)
+        .take(15)
+        .collect::<Vec<&str>>()
+        .join("\n");
+
+    // Replace the string DEFAULT_IMAGES = from the versions string
+    let versions = versions.replace("DEFAULT_IMAGES = ", "");
```
leovct, line 79: Same, as it might change...

crates/cdk/build.rs
vcastellm: yeah, it's the best solution we have to quickly solve the issue, this would be very fragile if done at runtime, but take into account two factors:

1. this will be run at buildtime, so no runtime errors will occur
2. It parses the json and outputs the json so it will 💥 at buildtime if anything changed and we'll detect it soon enough.

That said, there's a much comprehensive proposal than this command, but we've been requested to include this at this point.

crates/cdk/build.rs
vcastellm: Same response https://github.com/0xPolygon/cdk/pull/161#discussion_r1829669635

crates/cdk/versions.json
```diff
@@ -0,0 +1,15 @@
...
+  "zkevm_bridge_proxy_image": "haproxy:3.0-bookworm",
+  "zkevm_bridge_service_image": "hermeznetwork/zkevm-bridge-service:v0.6.0-RC1",
+  "zkevm_bridge_ui_image": "leovct/zkevm-bridge-ui:multi-network-2",
+  "zkevm_contracts_image": "leovct/zkevm-contracts:v8.0.0-rc.4-fork.12",
+  "zkevm_da_image": "0xpolygon/cdk-data-availability:0.0.10",
+  "zkevm_node_image": "hermeznetwork/zkevm-node:v0.7.3",
+  "zkevm_pool_manager_image": "hermeznetwork/zkevm-pool-manager:v0.1.1",
+  "zkevm_prover_image": "hermeznetwork/zkevm-prover:v8.0.0-RC14-fork.12",
+  "zkevm_sequence_sender_image": "hermeznetwork/zkevm-sequence-sender:v0.2.4"
+}
```
Stefan-Ethernal, line 15: Nitpick: add new line to the end.
```suggestion
}

```

crates/cdk/build.rs
Stefan-Ethernal, line 70: Seems like the comment is not up to date.

Stefan-Ethernal: I agree with @leovct, this should be improved in the (very near) future (if not through this PR) to rely on some regex to extract the part of the output we are interested in (`DEFAULT_IMAGES = { ... }`), as with this approach this is going to be deprecated and broken within a couple of days (IMO). :slightly_smiling_face:

rachit77 (approved):

PR #165, opened 2024-11-06
fix: Various pessimistic proofs fixes and adaption to `kurtosis-cdk`  pessimistic proof branch

goran-ethernal:
## Description

This PR includes:
1. Fix for replacing `InError` certificates. When a certificate is marked as `InError` on `agglayer`, we need to build a new one with the same height, that includes its bridges and claims, plus new ones that we saw in the meantime.
2. Fix for checking if `Candidate` and `Proven` certificates were also settled on the `agglayer`, because we cannot send a new one, until the previous one is either marked as `InError` or `Settled`.
3. Adapting `local_config` script to the new `kurtosis-cdk` version.
4. Fix where aggsender sent nil for imported bridge exits when it doesnt have any claims.
5. Fix where we didn't rollback `db` transaction in the the `aggsender_storage`.
6. Change db to store hashes as text
7. Add logs 

## Config changes
Parameter `SaveCertificatesToFiles` (bool) have been changed by `SaveCertificatesToFilesPath`: if is = "" is disabled, if not it write certificated in the specific path
```
[aggsender]
SaveCertificatesToFilesPath = ""
```


agglayer/client.go
```diff
@@ -91,6 +91,10 @@ func (c *AggLayerClient) WaitTxToBeMined(hash common.Hash, ctx context.Context)
 
 // SendCertificate sends a certificate to the AggLayer
 func (c *AggLayerClient) SendCertificate(certificate *SignedCertificate) (common.Hash, error) {
+	if err := defaultNilFieldsInCertificate(certificate); err != nil {
```
joanestebanr, line 94: You need to copy certificate before modified it

goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/165/commits/e2ff575c631a3a678617fe0172022ad631dae1f3

joanestebanr (approved):

agglayer/types.go
```diff
@@ -131,19 +153,54 @@ type SignedCertificate struct {
 	Signature *Signature `json:"signature"`
 }
 
+func (s *SignedCertificate) String() string {
+	return fmt.Sprintf("Certificate:%s, \nSignature: %s", s.Certificate.String(), s.Signature.String())
```
Stefan-Ethernal, line 157: Nitpick, in order to align the "labels":

```suggestion
	return fmt.Sprintf("Certificate:%s,\nSignature: %s", s.Certificate.String(), s.Signature.String())
```

aggsender/aggsender.go
```diff
@@ -63,6 +64,8 @@ func New(
 		return nil, err
 	}
 
+	logger.Info(cfg.String())
```
Stefan-Ethernal, line 67: WDYT: add some kind of label/prefix (e.g. `Aggsender config`) to this log?

Stefan-Ethernal (approved):

agglayer/types.go
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/165/commits/db602e6fdd89ecedb9aca8526e2b32ef12624a57

aggsender/aggsender.go
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/165/commits/db602e6fdd89ecedb9aca8526e2b32ef12624a57

agglayer/client.go
```diff
@@ -91,7 +91,9 @@ func (c *AggLayerClient) WaitTxToBeMined(hash common.Hash, ctx context.Context)
 
 // SendCertificate sends a certificate to the AggLayer
 func (c *AggLayerClient) SendCertificate(certificate *SignedCertificate) (common.Hash, error) {
-	response, err := rpc.JSONRPCCall(c.url, "interop_sendCertificate", certificate)
+	certificateToSend := certificate.CopyWithDefaulting()
```
vcastellm, line 94: Is this to the endpoint failing to receive nil values?

vcastellm (approved):
LGTM
G
T
M

agglayer/client.go
goran-ethernal: Yep. Even though we fixed in the aggsender code that we don't send nil values, for which we wrote a UT `TestSendCertificate_NoClaims`, we left this here as well as a fail safe.

PR #167, opened 2024-11-07
fix: acc input hash calculation

ToniRamirezM:
## Description

Fixes local acc input has calculation

### DIFF
--- a/aggregator/aggregator.go
+++ b/aggregator/aggregator.go
@@ -175,7 +175,7 @@ func New(
 		a.ctx, a.exit = context.WithCancel(a.ctx)
 	}
 
-	// Set function to handle the batches from the data stream
+	// Set function to handle events on L1
 	if !cfg.SyncModeOnlyEnabled {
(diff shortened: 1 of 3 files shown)


aggregator/aggregator.go
```diff
@@ -1481,13 +1514,12 @@ func (a *Aggregator) buildInputProver(
 		}
 	*/
 
-	a.accInputHashesMutex.Lock()
 	inputProver := &prover.StatelessInputProver{
 		PublicInputs: &prover.StatelessPublicInputs{
 			Witness: witness,
 			// Use calculated acc inputh hash as the RPC is not returning the correct one at the moment
 			// OldAccInputHash:   rpcOldBatch.AccInputHash().Bytes(),
```
vcastellm, line 1521: Remove comment?

vcastellm (approved):
LGTM in general but observe this lib that provides nice features and it's pretty well tested https://github.com/orcaman/concurrent-map

joanestebanr (approved):

PR #175, opened 2024-11-08
feat: add timestamps to certificate

goran-ethernal:
## Description

This PR adds three new fields to the `CertificateInfo` struct we save in `aggsender` `db`:
- `created_at` - a timestamp that indicates when was the certificate created on the `aggsender`.
- `updated_at` - a timestamp that indicates when was the certificate last updated. For example, it moved from Pending to Candidate.
- `signed_cerificated` - raw string of the certificate that we sent to `agglayer`.

All of these fields will be used for easier debugging and logging.

### DIFF
--- a/aggsender/db/migrations/0001.sql
+++ b/aggsender/db/migrations/0001.sql
@@ -10,6 +10,6 @@ CREATE TABLE certificate_info (
     from_block                  INTEGER NOT NULL,
     to_block                    INTEGER NOT NULL,
     created_at                  INTEGER NOT NULL,
-	updated_at                  INTEGER NOT NULL,
+    updated_at                  INTEGER NOT NULL,
     signed_certificate          TEXT
 );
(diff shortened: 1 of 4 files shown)


joanestebanr (approved):

aggsender/db/migrations/0001.sql
```diff
@@ -8,5 +8,8 @@ CREATE TABLE certificate_info (
     status                      INTEGER NOT NULL,
     new_local_exit_root         VARCHAR NOT NULL,
     from_block                  INTEGER NOT NULL,
-    to_block                    INTEGER NOT NULL
+    to_block                    INTEGER NOT NULL,
+    created_at                  INTEGER NOT NULL,
+	updated_at                  INTEGER NOT NULL,
```
Stefan-Ethernal, line 13: This one is misaligned (tab instead of 4 spaces):

```suggestion
    updated_at                  INTEGER NOT NULL,
```

Stefan-Ethernal (approved):

aggsender/db/migrations/0001.sql
goran-ethernal: fix: https://github.com/0xPolygon/cdk/pull/175/commits/f5f86ffc7c0eeae58e6a7c2b1979af6f06ca40cf

PR #178, opened 2024-11-12
feat: remove sanity check

ToniRamirezM:
## Description

Removes unneeded sanity check


joanestebanr (approved):
This sanity check is no longer required because the l1InfoRootIndex is comming from batchl2data and it's same place that is checked. The sanity  sanityCheck goal was to check that the batch data comming from Data-stream was right

PR #183, opened 2024-11-13
feat: aggsender e2e

joanestebanr:
## Description

Please include a summary of the change and which issue is fixed. Also, include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)


test/combinations/fork12-pessimistic.yml
```diff
@@ -0,0 +1,14 @@
+args:
+  agglayer_image: ghcr.io/agglayer/agglayer:0.2.0-rc.11
+  cdk_erigon_node_image: hermeznetwork/cdk-erigon:v2.60.0-beta8
+  cdk_node_image: cdk
```
goran-ethernal, line 4: Is this your local image?

test/scripts/agglayer_certificates_monitor.sh
```diff
@@ -0,0 +1,78 @@
...
+    current_time=$(date +%s)
+    if ((current_time > _end_time)); then
+        echo "[$(date '+%Y-%m-%d %H:%M:%S')] ❌ Exiting... Timeout reached not found the expected numbers of settled certs!"
+        exit 1
+    fi
+}
+
+###############################################################################
+function check_num_certificates(){
+    local _cmd="echo 'select status, count(*) from certificate_info group by status;' | sqlite3 /tmp/aggsender.sqlite"
```
goran-ethernal, line 33: Should we maybe parameterize the location of the `aggsender` db, and the `cdk-node-001` name? This is fine by me, just don't know if we parameterize this in other tests, as well as enclave name.

test/combinations/fork12-pessimistic.yml
joanestebanr: The e2e use a local docker build image

vcastellm (changes requested):
Blocking just to not forgot and merge with "jesteban/bump_aggsender"

vcastellm (changes requested):
Folder naming is inconsistent: fep vs pessimistic it should be `pp` + `fep` or `pessimistic` + `full-execution` or another option but consistent

Aside from that, to really be an e2e test, it should work the same way a user would do, querying the DB is something not realistic for the user. To check the certificates we should query the AggLayer using `getCertificateHeader` RPC 
 or `getLatestKnownCertificateHeader` that should be enough to know that the certificate was sent correctly.

test/scripts/agglayer_certificates_monitor.sh
```diff
@@ -0,0 +1,78 @@
...
+function check_timeout(){
+    local _end_time=$1
+    current_time=$(date +%s)
+    if ((current_time > _end_time)); then
+        echo "[$(date '+%Y-%m-%d %H:%M:%S')] ❌ Exiting... Timeout reached not found the expected numbers of settled certs!"
+        exit 1
+    fi
+}
+
+###############################################################################
```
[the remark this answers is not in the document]
joanestebanr: I like to separate the function on bash code  for better readability, but, anyway, I removed it

test/bats/fep/access-list-e2e.bats
```diff
@@ -1,6 +1,6 @@
 setup() {
-    load 'helpers/common-setup'
-    load 'helpers/common'
+    load '../../helpers/common-setup'
```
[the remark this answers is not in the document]
joanestebanr: This folder contains more files that are used on [unittest](https://github.com/0xPolygon/cdk/blob/d0e76a3d1361158aa24135f25d37ecc4af959755/aggoracle/e2e_test.go#L11)
For sure make sense, but in another PR to refactor this

.github/workflows/test-e2e.yml
```diff
@@ -70,7 +71,7 @@ jobs:
       with:
         repository: 0xPolygon/kurtosis-cdk
         path: "kurtosis-cdk"
-        ref: "v0.2.19"
+        ref: "jesteban/bump_aggsender"
```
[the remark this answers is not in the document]
joanestebanr: changer for v.0.20

vcastellm (changes requested):
> The test, as I explained, is a minium version and not ideal, for sure. The end-point required to retrieve a certificate is not released yet ( [check issue](https://github.com/agglayer/agglayer/issues/406)) , so the best approach is check on DB

I understand the situation, without those endpoints I agree checking the DB is the best way to check for certs, but the endpoints are released in 0.2.0-RC as I checked with the team, it should work with the image in this PR, the issue was closed a few mins ago, so let's not introduce more short-time debt here and work with what the agglayer have, no?

joanestebanr:
> > The test, as I explained, is a minium version and not ideal, for sure. The end-point required to retrieve a certificate is not released yet ( [check issue](https://github.com/agglayer/agglayer/issues/406)) , so the best approach is check on DB
> 
> I understand the situation, without those endpoints I agree checking the DB is the best way to check for certs, but the endpoints are released in 0.2.0-RC as I checked with the team, it should work with the image in this PR, the issue was closed a few mins ago, so let's not introduce more short-time debt here and work with what the agglayer have, no?


We have another PR to integrating this new end-point, so it make sense to merge this one as-is and in the new PR update the  `aggsender` code and also the e2e test to take advantatge of this

test/combinations/fork12-pessimistic.yml
```diff
@@ -0,0 +1,14 @@
...
+  zkevm_bridge_proxy_image: haproxy:3.0-bookworm
+  zkevm_bridge_service_image: hermeznetwork/zkevm-bridge-service:v0.6.0-RC1
+  zkevm_bridge_ui_image: leovct/zkevm-bridge-ui:multi-network
+  zkevm_contracts_image: nulyjkdhthz/zkevm-contracts:v9.0.0-rc.3-pp-fork.12
+  additional_services: []
+  consensus_contract_type: pessimistic
+  sequencer_type: erigon
+  erigon_strict_mode: false
+  zkevm_use_gas_token_contract: true
+  enable_normalcy: true
```
Stefan-Ethernal, line 14: Nitpick:

```suggestion
  enable_normalcy: true
  
```

test/bats/pp/bridge-e2e.bats
```diff
@@ -0,0 +1,76 @@
+setup() {
```
Stefan-Ethernal, line 1: Is it feasible to have the `bridge-e2e.bats` only in some common folder, if the bridge tests are the same for the FEP?

On the first glimpse, it seems like tests are (almost), if not the same. WDYT?

Stefan-Ethernal (approved):
Generally LGTM

PR #189, opened 2024-11-14
feat: sqlite aggregator

ToniRamirezM:
## Description

Use sqlite instead of postgresql in the aggregator.


test/config/test.config.toml
```diff
@@ -58,14 +58,7 @@ AggLayerURL = ""
...
 SequencerPrivateKey = {}
-	[Aggregator.DB]
-		Name = "aggregator_db"
-		User = "aggregator_user"
-		Password = "aggregator_password"
-		Host = "cdk-aggregator-db"
-		Port = "5432"
-		EnableLog = false	
-		MaxConns = 200
+DBPath = ""
```
joanestebanr, line 61: Setting to "" is not going to fail the execution?

joanestebanr (approved):

goran-ethernal (approved):

aggregator/db/migrations/0001.sql
```diff
@@ -1,32 +1,22 @@
 -- +migrate Down
```
Stefan-Ethernal, line 1: Should we drop `proof` and `sequence` tables here?

aggregator/db/dbstorage/proof.go
```diff
@@ -0,0 +1,356 @@
...
+	_, err := e.Exec(
+		addGeneratedProofSQL, proof.BatchNumber, proof.BatchNumberFinal, proof.Proof, proof.ProofID,
+		proof.InputProver, proof.Prover, proof.ProverID, generatingSince, createdAt, updatedAt,
+	)
+	return err
+}
+
+// UpdateGeneratedProof updates a generated proof in the storage
+func (d *DBStorage) UpdateGeneratedProof(ctx context.Context, proof *state.Proof, dbTx db.Txer) error {
+	const addGeneratedProofSQL = `
```
Stefan-Ethernal, line 260:
```suggestion
	const updateGeneratedProofSQL = `
```

aggregator/db/migrations/0001.sql
```diff
@@ -1,32 +1,22 @@
...
-DROP SCHEMA IF EXISTS aggregator CASCADE;
 
 -- +migrate Up
```
Stefan-Ethernal, line 3: Do we need a schema (e.g. named `aggregator`) as well, wdyt? Although the aggregator db is a separate file, so maybe it doesn't make sense at this point...

aggregator/db/migrations/0001.sql
```diff
@@ -1,32 +1,22 @@
...
-);
-
-CREATE TABLE IF NOT EXISTS aggregator.proof (
-	batch_num BIGINT NOT NULL REFERENCES aggregator.batch (batch_num) ON DELETE CASCADE,
 	batch_num_final BIGINT NOT NULL,
 	proof varchar NULL,
 	proof_id varchar NULL,
 	input_prover varchar NULL,
 	prover varchar NULL,
 	prover_id varchar NULL,
```
Stefan-Ethernal, line 11: Nitpick: it is probably more in a fashion of SQL Lite (https://www.sqlite.org/datatype3.html)
```suggestion
	proof TEXT NULL,
	proof_id TEXT NULL,
	input_prover TEXT NULL,
	prover TEXT NULL,
	prover_id TEXT NULL,
```

Stefan-Ethernal (approved):
LGTM, leaving some optional comments to consider.

aggregator/config.go
```diff
@@ -117,8 +108,8 @@ type Config struct {
 	// UseFullWitness is a flag to enable the use of full witness in the aggregator
 	UseFullWitness bool `mapstructure:"UseFullWitness"`
 
-	// DB is the database configuration
-	DB db.Config `mapstructure:"DB"`
+	// DBPath is the path to the database
+	DBPath string `mapstructure:"DBPath"`
```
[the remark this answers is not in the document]
ToniRamirezM: Well, I used DBPath just to keep the same name than on other components.

PR #191, opened 2024-11-15
fix: aggregating proofs

ToniRamirezM:
## Description

Fixes proofs aggregation.


joanestebanr (approved):

aggregator/aggregator.go
```diff
@@ -1163,21 +1163,43 @@ func (a *Aggregator) getAndLockBatchToProve(
 		virtualBatch.L1InfoRoot = &l1InfoRoot
 	}
 
+	// Ensure the old acc input hash is in memory
+	oldAccInputHash := a.getAccInputHash(batchNumberToVerify - 1)
+	if oldAccInputHash == (common.Hash{}) && batchNumberToVerify > 1 {
+		tmpLogger.Warnf("AccInputHash for batch - 1 (%d) is not in memory. Waiting ...", batchNumberToVerify-1)
```
Stefan-Ethernal, line 1169: WDYT?
```suggestion
		tmpLogger.Warnf("AccInputHash for the previous batch (%d) is not in memory. Waiting ...", batchNumberToVerify-1)
```

aggregator/aggregator.go
```diff
@@ -1346,6 +1368,20 @@ func (a *Aggregator) tryGenerateBatchProof(ctx context.Context, prover ProverInt
...
+	// Sanity Check: acc input hash from the proof must match the one from the batch
+	if a.cfg.BatchProofSanityCheckEnabled && (accInputHash != common.Hash{}) &&
+		(accInputHash != batchToProve.AccInputHash) {
+		for {
+			tmpLogger.Errorf("Acc input hash from the proof does not match the expected for "+
+				"batch %d: Proof = [%s] Expected = [%s]",
+				batchToProve.BatchNumber, accInputHash.String(), batchToProve.AccInputHash.String(),
+			)
+			time.Sleep(a.cfg.RetryTime.Duration)
+		}
```
Stefan-Ethernal, line 1380: If this is the halting logic, should we annotate in the logs that we are halting the aggregator?

Asking, because it seems to me that neither `batchToProve` or accInputHash is rewritten in the for loop and from what I can see, there is no other goroutine that would update these two vars.

vcastellm (approved):
LGTM

PR #197, opened 2024-11-22
feat: remove local `db` from `aggsender`

goran-ethernal:
## Description

This PR removes the local `aggsender` `db`, and uses in memory saving of the last sent certificate.

On startup, the `aggsender` will get the last sent certificate for its network by using the `interop_getLatestKnownCertificateHeader` rpc endpoint on `agglayer`. If it is `nil`, it means that `agglayer` doesn't have any certificates from our network. If it has, we will save it in memory as the last sent certificate.

Then, the flow is as it was:
- we wait for the moment in the L1 epoch to build a new certificate.
- if the last sent certificate is nil (`agglayer` has no certificates from our network), or if it's status is `Settled` or `InError`, `aggsender` will proceed to build a new certificate.
- if the last sent certificate has any other status than `Settled` or `InError` it means we can not build a new certificate, so we wait for the next chance to build it.

Pros of removing local `aggsender` `db`:
- we are relaying solely on `agglayer` as a source of truth.
- code is less complex.
- `aggsender` config is smaller.
- `aggsender` is faster since it saves the last sent certificate in memory, and it doesn't need to query `db`.

Cons:
- No history tracking.
- Some situations in production can be more difficult to track. e.g: there are 2 `aggsenders` sending certs with same `NetworkID`, or someone was manually sending certificates to the `agglayer`.


agglayer/client.go
```diff
@@ -158,3 +159,24 @@ func (c *AggLayerClient) GetEpochConfiguration() (*ClockConfiguration, error) {
 
 	return result, nil
 }
+
+// GetLatestKnownCertificateHeader returns the last certificate header submitted by networkID
+func (c *AggLayerClient) GetLatestKnownCertificateHeader(networkID uint32) (*CertificateHeader, error) {
```
vcastellm, line 164: Already implemented in https://github.com/0xPolygon/cdk/pull/192/files#diff-1ebfd4f92574f105b4c8a06e8e5ab994b73d7dcd338971706d5cc2e2b4d577e2?

vcastellm (commented):
Rebase to deduce diff?

goran-ethernal:
I am closing this for now, as per agreement with the team.

PR #199, opened 2024-11-25
fix: cdk603 error calculating previousLocalExitRoot

joanestebanr:
## Description

Since `agglayer:0.2.0-rc.15` the responses of ` CertificateHeader` add a new field: `PreviousLocalExitRoot`
- Fixes error calculating  previousLocalExitRoot on new certificate if a certificate is inError

NOTE: **Database incompatibility with previous version!, require to delete it!**

### DIFF
--- a/agglayer/client_test.go
+++ b/agglayer/client_test.go
@@ -45,7 +45,7 @@ func TestExploratoryGetEpochConfiguration(t *testing.T) {
 
 func TestExploratoryGetLatestKnownCertificateHeader(t *testing.T) {
 	t.Skip("This test is exploratory and should be skipped")
-	aggLayerClient := NewAggLayerClient("http://localhost:32781")
+	aggLayerClient := NewAggLayerClient("http://localhost:32843")
 	cert, err := aggLayerClient.GetLatestKnownCertificateHeader(1)
 	require.NoError(t, err)
 	fmt.Print(cert)
@@ -116,7 +116,9 @@ func TestGetLatestKnownCertificateHeaderOkResponse(t *testing.T) {
 	cert, err := sut.GetLatestKnownCertificateHeader(1)
 	require.NotNil(t, cert)
 	require.NoError(t, err)
+	require.Nil(t, cert.PreviousLocalExitRoot)
 }
+
 func TestGetLatestKnownCertificateHeaderErrorResponse(t *testing.T) {
 	sut := NewAggLayerClient(testURL)
 	jSONRPCCall = func(url, method string, params ...interface{}) (rpc.Response, error) {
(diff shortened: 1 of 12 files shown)


ToniRamirezM (approved):

agglayer/client_test.go
```diff
@@ -45,7 +45,7 @@ func TestExploratoryGetEpochConfiguration(t *testing.T) {
 
 func TestExploratoryGetLatestKnownCertificateHeader(t *testing.T) {
 	t.Skip("This test is exploratory and should be skipped")
-	aggLayerClient := NewAggLayerClient("http://localhost:32781")
+	aggLayerClient := NewAggLayerClient("http://localhost:32843")
```
vcastellm, line 48: Will this keep changing? should we move this and the previous ones to a constant?

agglayer/types.go
```diff
@@ -574,36 +576,41 @@ func (p *GenericPPError) String() string {
...
+		return nilStr
 	}
 	errors := ""
 	if c.Error != nil {
 		errors = c.Error.String()
 	}
-
-	return fmt.Sprintf("Height: %d, CertificateID: %s, NewLocalExitRoot: %s. Status: %s. Errors: [%s]",
-		c.Height, c.CertificateID.String(), c.NewLocalExitRoot.String(), c.Status.String(), errors)
+	previousLocalExitRoot := "nil"
```
vcastellm, line 607: use the `nilStr` constant?

agglayer/types.go
```diff
@@ -23,6 +23,8 @@ const (
 	Candidate
 	InError
 	Settled
+
+	nilStr = "nil"
```
vcastellm, line 27: Why is this needed, agglayer API?

aggsender/aggsender_test.go
```diff
@@ -283,7 +285,7 @@ func TestAggSenderStart(t *testing.T) {
 		ctx,
 		log.WithFields("test", "unittest"),
 		Config{
-			StoragePath:          "file::memory:?cache=shared",
+			StoragePath:          "file:TestAggSenderStart?mode=memory&cache=shared",
```
vcastellm, line 288: Are we 100% sure this is using memory or a file with this name?

db/meddler.go
```diff
@@ -163,6 +163,16 @@ func (b HashMeddler) PostRead(fieldPtr, scanTarget interface{}) error {
 	}
 	field, ok := fieldPtr.(*common.Hash)
 	if !ok {
+		hashPtr, ok := fieldPtr.(**common.Hash)
```
vcastellm, line 166: In which case this can be a pointer to the pointer?

Stefan-Ethernal: If I'm not mistaken, this can be the case when the hash value is `NULL` in the DB.

agglayer/client_test.go
Stefan-Ethernal: I believe so... Every time Kurtosis is spun up, it assigns different port values. Anyway, this test is skipped, as it is only for some debug/exploratory purposes (and therefore we might want to remove it at some point once the agglayer becomes stable enough).

PR #200, opened 2024-11-26
feat(aggsender): add more data to certificate metadata

vcastellm:
In this PR we're introducing a new struct `CertificateMetadata` to represent the data serialized and deserialized to the `Certificate.metadata` field.

The new data, `FromBlock` and `CreatedAt` is stored in the hash as bigints.

NOTE: This PR doesn't take into account backwards compat. We should rethink if this needs to be considered.


Stefan-Ethernal (approved):

aggsender/aggsender.go
```diff
@@ -723,28 +725,49 @@ func extractSignatureData(signature []byte) (r, s common.Hash, isOddParity bool,
 }
 
 // createCertificateMetadata creates a certificate metadata from given input
-func createCertificateMetadata(toBlock uint64) common.Hash {
-	return common.BigToHash(new(big.Int).SetUint64(toBlock))
+func createCertificateMetadata(fromBlock, toBlock, createdAt uint64) common.Hash {
```
joanestebanr, line 728: Maybe as parameter you can use `types.CertificateMetadata`

aggsender/aggsender.go
```diff
@@ -723,28 +725,49 @@ func extractSignatureData(signature []byte) (r, s common.Hash, isOddParity bool,
...
 		NewLocalExitRoot:  c.NewLocalExitRoot,
-		FromBlock:         0,
-		ToBlock:           extractFromCertificateMetadataToBlock(c.Metadata),
+		FromBlock:         meta.FromBlock,
+		ToBlock:           meta.ToBlock,
 		Status:            c.Status,
-		CreatedAt:         now,
-		UpdatedAt:         now,
+		CreatedAt:         meta.CreatedAt,
+		UpdatedAt:         meta.CreatedAt,
```
joanestebanr, line 770: I think that this must be now

joanestebanr (changes requested):
Please, fill the PR description

vcastellm:
> Please, fill the PR description

thx, forgot about it when moving out of draft, done.

aggsender/aggsender.go
vcastellm: that's one possibility I considered but now that you mention it can be even better abstracted, see https://github.com/0xPolygon/cdk/pull/200/commits/b7950e906992bc8805cb04676bcbffaf4e2fbb7b wdyt?

aggsender/aggsender.go
joanestebanr: Yes, it is. The `updateAt` is store the moment that `aggsender` write some update on this row, that is the idea.

vcastellm: makes sense

joanestebanr (approved):

PR #202, opened 2024-11-27
chore: bump kurtosis-cdk version to support pre-deployed gas token

leovct:
## Description

Related to https://github.com/0xPolygon/kurtosis-cdk/pull/386

It looks like the CI doesn't run the e2e tests on my PR, maybe because it comes from an external fork? I've created a [local PR](https://github.com/leovct/cdk/pull/1) on my fork to make sure they all pass. Note that the fork12-pp test doesn't pass on the `develop` branch either.

Screenshot 2024-11-27 at 10 48 44

PR #208, opened 2024-11-28
feat: write on database the number of retries per certificate and the certificates in a history table

joanestebanr:
## Description

- Add new field `retries` to database, that keep the count of times of regenerated the certificate
- The discarded certificates are move (if configuration allow that) to a new table `certificate_info_history`
- Cherry-picked #202 to fix e2e-test

## Configuration
```
[AggSender]
KeepCertificatesHistory = true
```
## Next steps
- Define the retain policy

### DIFF
--- a/aggsender/config.go
+++ b/aggsender/config.go
@@ -36,6 +36,8 @@ type Config struct {
 	// DelayBeetweenRetries is the delay between retries:
 	//  is used on store Certificate and also in initial check
 	DelayBeetweenRetries types.Duration `mapstructure:"DelayBeetweenRetries"`
+	// KeepCertificatesHistory is a flag to keep the certificates history on storage
+	KeepCertificatesHistory bool `mapstructure:"KeepCertificatesHistory"`
 }
 
 // String returns a string representation of the Config
--- a/aggsender/db/aggsender_db_storage.go
+++ b/aggsender/db/aggsender_db_storage.go
@@ -36,26 +36,33 @@ type AggSenderStorage interface {
 
 var _ AggSenderStorage = (*AggSenderSQLStorage)(nil)
 
+// AggSenderSQLStorageConfig is the configuration for the AggSenderSQLStorage
+type AggSenderSQLStorageConfig struct {
+	DBPath                  string
+	KeepCertificatesHistory bool
+}
+
 // AggSenderSQLStorage is the struct that implements the AggSenderStorage interface
 type AggSenderSQLStorage struct {
 	logger *log.Logger
 	db     *sql.DB
+	cfg    AggSenderSQLStorageConfig
 }
 
 // NewAggSenderSQLStorage creates a new AggSenderSQLStorage
-func NewAggSenderSQLStorage(logger *log.Logger, dbPath string) (*AggSenderSQLStorage, error) {
-	if err := migrations.RunMigrations(dbPath); err != nil {
+func NewAggSenderSQLStorage(logger *log.Logger, cfg AggSenderSQLStorageConfig) (*AggSenderSQLStorage, error) {
+	db, err := db.NewSQLiteDB(cfg.DBPath)
+	if err != nil {
 		return nil, err
 	}
-
-	db, err := db.NewSQLiteDB(dbPath)
-	if err != nil {
+	if err := migrations.RunMigrations(logger, db); err != nil {
 		return nil, err
 	}
 
 	return &AggSenderSQLStorage{
 		db:     db,
 		logger: logger,
+		cfg:    cfg,
 	}, nil
 }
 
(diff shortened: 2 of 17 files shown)


aggsender/db/aggsender_db_storage.go
```diff
@@ -205,8 +228,8 @@ func (a *AggSenderSQLStorage) UpdateCertificate(ctx context.Context, certificate
 		}
 	}()
 
-	if _, err = tx.Exec(`UPDATE certificate_info SET status = $1 WHERE certificate_id = $2;`,
-		certificate.Status, certificate.CertificateID.String()); err != nil {
+	if _, err = tx.Exec(`UPDATE certificate_info SET status = $1, updated_at=$2 WHERE certificate_id = $3;`,
```
ToniRamirezM, line 231:
```suggestion
	if _, err = tx.Exec(`UPDATE certificate_info SET status = $1, updated_at = $2 WHERE certificate_id = $3;`,
```

ToniRamirezM (approved):

aggsender/config.go
```diff
@@ -36,6 +36,8 @@ type Config struct {
 	// DelayBeetweenRetries is the delay between retries:
 	//  is used on store Certificate and also in initial check
 	DelayBeetweenRetries types.Duration `mapstructure:"DelayBeetweenRetries"`
+	// KeepCertificatesHistory is a flag to keep the certificates history on storage
+	KeepCertificatesHistory bool `mapstructure:"KeepCertificatesHistory"`
```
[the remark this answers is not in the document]
joanestebanr: store the history of certificates on DB, if it's `false` the certificate not used are discarded

joanestebanr:
> What's the difference between `certificate_info_history` and `certificate_info`? Could just the `certificate_info` table contain all records and filter by status? both table have the same data model

The idea is that `certificate_info_history` could be empty if `KeepCertificatesHistory` is `false`.
The table `certificate_info`  enforce the integrity of data having as primary key `height` but the table with history is `height` and `retry`

aggsender/db/aggsender_db_storage.go
```diff
@@ -147,14 +154,30 @@ func (a *AggSenderSQLStorage) SaveLastSentCertificate(ctx context.Context, certi
...
+		return fmt.Errorf("saveLastSentCertificate commit. Err: %w", err)
 	}
 
 	a.logger.Debugf("inserted certificate - Height: %d. Hash: %s", certificate.Height, certificate.CertificateID)
 
 	return nil
 }
 
+func (a *AggSenderSQLStorage) moveCertificateToHistory(tx db.Querier, certificate *types.CertificateInfo) error {
+	if a.cfg.KeepCertificatesHistory {
```
vcastellm, line 166: the method is called `moveCertificateToHistory` semantically it implies that it will move the cert to history, but it won't depending on the conf, so better move the check outside the function? or change the method name to `maybeMoveCertificateToHistory`

aggsender/db/aggsender_db_storage.go
```diff
@@ -147,14 +154,30 @@ func (a *AggSenderSQLStorage) SaveLastSentCertificate(ctx context.Context, certi
...
-		return err
+		return fmt.Errorf("saveLastSentCertificate commit. Err: %w", err)
 	}
 
 	a.logger.Debugf("inserted certificate - Height: %d. Hash: %s", certificate.Height, certificate.CertificateID)
 
 	return nil
 }
 
+func (a *AggSenderSQLStorage) moveCertificateToHistory(tx db.Querier, certificate *types.CertificateInfo) error {
```
goran-ethernal:
```suggestion
func (a *AggSenderSQLStorage) archiveOrRemoveCertificate(tx db.Querier, certificate *types.CertificateInfo) error {
```

aggsender/db/aggsender_db_storage.go
joanestebanr: I change for `moveCertificateToHistoryOrDelete`

aggsender/db/migrations/0001.sql
```diff
@@ -1,16 +1,35 @@
...
+    certificate_id              VARCHAR NOT NULL,
+    status                      INTEGER NOT NULL,
+    previous_local_exit_root    VARCHAR,
+    new_local_exit_root         VARCHAR NOT NULL,
+    from_block                  INTEGER NOT NULL,
+    to_block                    INTEGER NOT NULL,
+    created_at                  INTEGER NOT NULL,
+    updated_at                  INTEGER NOT NULL,
+    signed_certificate          TEXT,
+    PRIMARY KEY (height, retry_count)
```
goran-ethernal, line 34: This can also conflict with a certificate that was retried before. Maybe use height and certificate id as keys?

vcastellm (approved):
LGTM aside from minor comment

aggsender/db/migrations/0001.sql
joanestebanr: It never have to conflict, if it conflicts means that we have something wrong.

aggsender/db/aggsender_db_storage.go
joanestebanr: I have already renamed to `moveCertificateToHistoryOrDelete` because a PR comment, do you think that is good enough?

goran-ethernal: That is fine.

goran-ethernal (approved):

PR #216, opened 2024-12-03
refactor: encode metadata to be backwards compatible

vcastellm:
Refactoring the encoding of the values into the metadata field makes it compatible with the previously existing values, added a test for that.

Also moving tests to the types package where it belongs.

### DIFF
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -195,6 +195,7 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayer.SignedCertif
 		ToBlock:   toBlock,
 		Bridges:   bridges,
 		Claims:    claims,
+		CreatedAt: uint32(time.Now().UTC().Unix()),
 	}
 
 	certificateParams, err = a.limitCertSize(certificateParams)
@@ -228,7 +229,6 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayer.SignedCertif
 		return nil, fmt.Errorf("error marshalling signed certificate. Cert:%s. Err: %w", signedCertificate.Brief(), err)
 	}
 
-	createdTime := time.Now().UTC().UnixMilli()
 	prevLER := common.BytesToHash(certificate.PrevLocalExitRoot[:])
 	certInfo := types.CertificateInfo{
 		Height:                certificate.Height,
@@ -238,8 +238,8 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayer.SignedCertif
 		PreviousLocalExitRoot: &prevLER,
 		FromBlock:             fromBlock,
 		ToBlock:               toBlock,
-		CreatedAt:             createdTime,
-		UpdatedAt:             createdTime,
+		CreatedAt:             int64(certificateParams.CreatedAt),
+		UpdatedAt:             int64(certificateParams.CreatedAt),
 		SignedCertificate:     string(raw),
 	}
 	// TODO: Improve this case, if a cert is not save in the storage, we are going to settle a unknown certificate
@@ -396,7 +396,7 @@ func (a *AggSender) buildCertificate(ctx context.Context,
 	meta := types.NewCertificateMetadata(
 		certParams.FromBlock,
 		uint32(certParams.ToBlock-certParams.FromBlock),
-		uint32(time.Now().UTC().Unix()),
+		certParams.CreatedAt,
 	)
 
 	return &agglayer.Certificate{
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -381,7 +381,7 @@ func (a *AggSender) buildCertificate(ctx context.Context,
 		return nil, fmt.Errorf("error getting imported bridge exits: %w", err)
 	}
 
-	depositCount := certParams.MaxDepoitCount()
+	depositCount := certParams.MaxDepositCount()
 
 	exitRoot, err := a.l2Syncer.GetExitRootByIndex(ctx, depositCount)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


aggsender/types/types.go
```diff
@@ -131,7 +131,7 @@ type CertificateMetadata struct {
 }
 
 // NewCertificateMetadataFromHash returns a new CertificateMetadata from the given hash
-func NewCertificateMetadata(fromBlock, toBlock, createdAt uint64) *CertificateMetadata {
+func NewCertificateMetadata(toBlock, fromBlock, createdAt uint64) *CertificateMetadata {
```
Stefan-Ethernal, line 134: IMO it is more intuitive to provide the `fromBlock` as the first parameter (although we are encoding the `toBlock` first). WDYT?

```suggestion
func NewCertificateMetadata(fromBlock, toBlock, createdAt uint64) *CertificateMetadata {
```

aggsender/types/types.go
```diff
@@ -144,24 +144,24 @@ func NewCertificateMetadataFromHash(hash common.Hash) *CertificateMetadata {
...
 }
 
 // ToHash returns the hash of the metadata
 func (c *CertificateMetadata) ToHash() common.Hash {
 	b := make([]byte, common.HashLength) // 32-byte hash
 
-	// Encode fromBlock into first 8 bytes
-	binary.BigEndian.PutUint64(b[0:8], c.FromBlock)
+	// Encode createdAt
+	binary.BigEndian.PutUint64(b[8:16], c.CreatedAt)
```
Stefan-Ethernal, line 158: Maybe add some comment, that prior to #200, we were encoding just the `ToBlock` and that's why we are skipping the first 8 bytes...?

Although _probably_ it is worth considering versioning metadata (add the version field and act accordingly). WDYT?

vcastellm: ok to 1.

Versioning doesn't makes much sense to me as there's only 8 bytes of space left, if we add versioning we'll use all space, no space for more changes. I think it's worth handling changes here and keeping those 8 bytes free for the future.

aggsender/types/types.go
```diff
@@ -144,24 +144,29 @@ func NewCertificateMetadataFromHash(hash common.Hash) *CertificateMetadata {
 	b := hash.Bytes()
 
 	return NewCertificateMetadata(
-		binary.BigEndian.Uint64(b[0:8]),
-		binary.BigEndian.Uint64(b[8:16]),
 		binary.BigEndian.Uint64(b[16:24]),
+		binary.BigEndian.Uint64(b[24:32]),
+		binary.BigEndian.Uint64(b[8:16]),
 	)
```
Stefan-Ethernal, line 150: Shouldn't this function decode the certificate metadata with the old format (namely, support the format where the `toBlock` field occupied the first 8 bytes)?

joanestebanr: I think that requires to set the 'createAt` to `now` if it's the previous metadata version

joanestebanr (changes requested):
Please, merge it at `release/v0.5.0`

aggsender/types/types.go
joanestebanr: I agree with @Stefan-Ethernal , we need to versioning metadata... We need to think in a way of doing this

vcastellm:
> Please, merge it at `release/v0.5.0`

Instead of rebasing, let's merge it to `develop` then I'll cherry-pick to `v0.5.0`

aggsender/types/types.go
vcastellm: This is decoding @joanestebanr, so we need to take what's in `hash`.

aggsender/types/types.go
vcastellm: Do you prefer to ocuppy the remaining 8 bytes with the version?

vcastellm:
Rebased onto `release/v0.5.0` @joanestebanr

aggsender/aggsender.go
```diff
@@ -769,28 +774,25 @@ func extractSignatureData(signature []byte) (r, s common.Hash, isOddParity bool,
...
-
 func NewCertificateInfoFromAgglayerCertHeader(c *agglayer.CertificateHeader) *types.CertificateInfo {
 	if c == nil {
 		return nil
 	}
 	now := time.Now().UTC().UnixMilli()
+	meta := types.NewCertificateMetadataFromHash(c.Metadata)
+	toBlock := meta.FromBlock + uint64(meta.Offset)
+	if meta.Version < 1 {
+		toBlock = meta.ToBlock
```
joanestebanr, line 787: The `createdAt` need to be set to `now`

aggsender/aggsender.go
```diff
@@ -769,28 +774,25 @@ func extractSignatureData(signature []byte) (r, s common.Hash, isOddParity bool,
...
 		Height:            c.Height,
 		CertificateID:     c.CertificateID,
 		NewLocalExitRoot:  c.NewLocalExitRoot,
-		FromBlock:         0,
-		ToBlock:           extractFromCertificateMetadataToBlock(c.Metadata),
+		FromBlock:         meta.FromBlock,
+		ToBlock:           toBlock,
 		Status:            c.Status,
-		CreatedAt:         now,
+		CreatedAt:         int64(meta.CreatedAt),
```
joanestebanr: maybe make sense to change the type of `CertificateInfo` for tstamp to `uint32` all of them, WDYT?

joanestebanr (approved):

aggsender/aggsender.go
vcastellm: agree

Stefan-Ethernal (approved):

PR #217, opened 2024-12-03
feat: limit cert by estimated size

joanestebanr:
## Description

This set a limit on certificate size that can be set on config file (the size of certificate it's an estimation)
There are a special situation that is going to send a certificate bigger than maximium: 
- [ initial_block, end_block ] -> size exceed
- [ initial_block, (end_block -1) ] -> no bridges, so we have to send previous one even that exceed the maximum size

## Config
MaxCertSize: max size in bytes of certificate. Default 8Mb. Currently the maximum on Agglayer is 10Mb
```
[AggSender]
MaxCertSize = 8388608
```


ToniRamirezM (approved):

aggsender/aggsender.go
```diff
@@ -187,12 +187,12 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayer.SignedCertif
 		return nil, nil
 	}
 
-	claims, err := a.l2Syncer.GetClaims(ctx, fromBlock, toBlock)
+	toBlock, bridges, claims, err := a.limitCertSize(ctx, fromBlock, toBlock)
```
goran-ethernal, line 190: Should we remove the line above this that gets bridges published. And just check the array of bridges that `limitCertSize` returns?

aggsender/aggsender.go
```diff
@@ -265,6 +265,46 @@ func (a *AggSender) saveCertificateToStorage(ctx context.Context, cert types.Cer
 	return nil
 }
 
+const (
+	estimatedSizeBridgeExit = 250
+	estimatedSizeClaim      = 44000
+)
+
+func (a *AggSender) limitCertSize(ctx context.Context, fromBlock, toBlock uint64) (uint64,
```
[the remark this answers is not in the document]
joanestebanr: good catch!

aggsender/aggsender.go
```diff
@@ -265,6 +265,46 @@ func (a *AggSender) saveCertificateToStorage(ctx context.Context, cert types.Cer
...
+	[]bridgesync.Bridge, []bridgesync.Claim, error) {
+	var err error
+	var bridges, previousBridges []bridgesync.Bridge
+	var claims, previousClaims []bridgesync.Claim
+	for true {
```
vcastellm, line 278:
```suggestion
	for {
```

aggsender/aggsender.go
joanestebanr: I think about that, I discarded because it's going  to  over-complicate the function limitCertSize because it have to detect that is the first step, and return if no bridges.

goran-ethernal: I think you are already something like that in the `limit` function:

```go
if len(bridges) == 0 {
	// We can't reduce more the certificate, so this is the minium size
	a.log.Warnf("We reach the minium size of bridge: %d that is bigger than the max size of certificate: %d",
		a.estimateSizeCert(len(previousBridges), len(previousClaims)), a.cfg.MaxCertSize)
	return toBlock + 1, previousBridges, previousClaims, nil
}
```

I think this is enough and correct, so we can remove what I suggested. If it is the first step, it will just get the bridges of initial `from` and `to` and if there is no bridges, it will simply return empty array. Then it will be ok to check after the call of the `limit` function.

vcastellm (commented):
LGTM but I understand that the size is not 100% accurate but based on estimations right?

Another issue I see is that the user can overflow the agglayer, should we limit the cert size to AggLayer max limit?

aggsender/types/certificate_params.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/0xPolygon/cdk/bridgesync"
+)
+
+const (
+	EstimatedSizeBridgeExit = 250
+	EstimatedSizeClaim      = 44000
+)
+
+type CertificateLocalParams struct {
```
goran-ethernal, line 14: Add comments to the struct and functions.

aggsender/aggsender_test.go
```diff
@@ -1934,6 +1989,31 @@ type aggsenderTestData struct {
 	testCerts            []aggsendertypes.CertificateInfo
 }
 
+func NewBridgesData(num int, blockNum []uint64) []bridgesync.Bridge {
```
goran-ethernal, line 1992: Pass `t *testin.T` to the function and add the `t.Helper()` line to it to mark it as test function, so that we don't use it anywhere else, and that sonar marks it as test and skip it in its analysis.

aggsender/aggsender_test.go
```diff
@@ -1934,6 +1989,31 @@ type aggsenderTestData struct {
...
+			BlockNum:      blockNum[i%len(blockNum)],
+			BlockPos:      0,
+			LeafType:      agglayer.LeafTypeAsset.Uint8(),
+			OriginNetwork: 1,
+		})
+	}
+	return res
+}
+
+func NewClaimData(num int, blockNum []uint64) []bridgesync.Claim {
```
goran-ethernal, line 2005: Pass `t *testin.T` to the function and add the `t.Helper()` line to it to mark it as test function, so that we don't use it anywhere else, and that sonar marks it as test and skip it in its analysis.

aggsender/types/certificate_params.go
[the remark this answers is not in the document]
joanestebanr: I'll rename to `CertificateBuildParams`

PR #219, opened 2024-12-03
chore: release v0.4.1

vcastellm:
Release v0.4.1


ToniRamirezM (approved):

rachit77 (approved):

joanestebanr (changes requested):
Why the PR is not passing Unit test?

vcastellm:
> Why the PR is not passing Unit test?

It's explained in the slack message

joanestebanr:
> > Why the PR is not passing Unit test?
> 
> It's explained in the slack message: upgrades in Kurtosis that broke the testing

Just for clarify: the unittest are not using Kurtosis, so the problem is another.

joanestebanr (approved):
I suggest to fix unittest before merging

PR #225, opened 2024-12-04
test: e2e checking for multiples certificates submitted

rachit77:
## Description

Task: https://polygon.atlassian.net/browse/CDK-598


goran-ethernal (approved):

test/bats/pp/bridge-ci-e2e.bats
```diff
@@ -0,0 +1,100 @@
+setup() {
```
Stefan-Ethernal: Why not expanding the existing one to send multiple deposits? Or even better to somehow parameterize how many of deposits we want to make?

So the optimal approach for me is to keep only the `bridge-e2e.bats` and either change the existing `Native gas token deposit to WETH` test to perform multiple deposits, or parameterize that test with deposits count (so that we can provide only single deposit in case we want to debug it for example). WDYT?

PR #228, opened 2024-12-04
feat: allow metadata field of ImportedBridges as hash

joanestebanr:
## Description

The certificate field `metadata`  of `BridgeExit` and `ImportedBridgeExit` instead of containing the full metadata are going to have the hash if length > 0. 
This behevaiour can be configured using a new config param `AggSender.BridgeMetadataAsHash` (by default: `true`)

Other fixes: 
- Missing missing default values for `L1InfoTreeSync.RetryAfterErrorPeriod` and `MaxRetryAttemptsAfterError`
- Improved estimateSize of certificate
- Added e2e bridge messages but disabled because it keep failing (maybe we neeed another PR to fix that)


aggsender/aggsender.go
```diff
@@ -411,6 +411,15 @@ func (a *AggSender) convertClaimToImportedBridgeExit(claim bridgesync.Claim) (*a
 	if claim.IsMessage {
 		leafType = agglayer.LeafTypeMessage
 	}
+	var metaData []byte
+	var MetadataIsHashed bool
+	if a.cfg.ImportedBridgeMetadataAsHash && len(claim.Metadata) > 0 {
```
goran-ethernal: If `metadata` is empty, should we hash it as zero hash if `ImportedBridgeMetadataAsHash`? Maybe we should check this with the `agglayer` team.

aggsender/aggsender.go
```diff
@@ -444,6 +454,16 @@ func (a *AggSender) getBridgeExits(bridges []bridgesync.Bridge) []*agglayer.Brid
 	bridgeExits := make([]*agglayer.BridgeExit, 0, len(bridges))
 
 	for _, bridge := range bridges {
+		var metaData []byte
+		var MetadataIsHashed bool
+		if a.cfg.ImportedBridgeMetadataAsHash && len(bridge.Metadata) > 0 {
```
goran-ethernal: Maybe we can have this as a separate function, since it is the same code for bridge exits and bridge exits in a claim.

aggsender/aggsender.go
joanestebanr: We agree with  Simon to set empty (to reduce the size of transfer leafs)

aggsender/aggsender.go
```diff
@@ -405,21 +405,28 @@ func (a *AggSender) buildCertificate(ctx context.Context,
...
-func (a *AggSender) convertClaimToImportedBridgeExit(claim bridgesync.Claim) (*agglayer.ImportedBridgeExit, error) {
-	leafType := agglayer.LeafTypeAsset
-	if claim.IsMessage {
-		leafType = agglayer.LeafTypeMessage
-	}
+// createCertificateMetadata creates the metadata for the certificate
+// it returns: newMetadata + bool if the metadata is hashed or not
+func convertBridgeMetadata(metadata []byte, importedBridgeMetadataAsHash bool) ([]byte, bool) {
 	var metaData []byte
 	var MetadataIsHashed bool
```
goran-ethernal:
```suggestion
	var isMetadataHashed bool
```

aggsender/aggsender.go
```diff
@@ -405,21 +405,28 @@ func (a *AggSender) buildCertificate(ctx context.Context,
...
+	return metaData, MetadataIsHashed
+}
+
+// convertClaimToImportedBridgeExit converts a claim to an ImportedBridgeExit object
+func (a *AggSender) convertClaimToImportedBridgeExit(claim bridgesync.Claim) (*agglayer.ImportedBridgeExit, error) {
+	leafType := agglayer.LeafTypeAsset
+	if claim.IsMessage {
+		leafType = agglayer.LeafTypeMessage
+	}
+	metaData, MetadataIsHashed := convertBridgeMetadata(claim.Metadata, a.cfg.ImportedBridgeMetadataAsHash)
```
goran-ethernal:
```suggestion
	metaData, isMetadataHashed := convertBridgeMetadata(claim.Metadata, a.cfg.ImportedBridgeMetadataAsHash)
```

aggsender/config.go
```diff
@@ -37,6 +37,8 @@ type Config struct {
 	// MaxCertSize is the maximum size of the certificate (the emitted certificate can be bigger that this size)
 	// 0 is infinite
 	MaxCertSize uint `mapstructure:"MaxCertSize"`
+	// ImportedBridgeMetadataAsHash is a flag to import the bridge metadata as hash
+	ImportedBridgeMetadataAsHash bool `mapstructure:"ImportedBridgeMetadataAsHash"`
```
goran-ethernal: Since it is used for both bridge exits and imported bridge exits, not just the imported exits.

```suggestion
	BridgeMetadataAsHash bool `mapstructure:"ImportedBridgeMetadataAsHash"`
```

agglayer/types.go
```diff
@@ -267,6 +267,7 @@ type BridgeExit struct {
 	DestinationNetwork uint32         `json:"dest_network"`
 	DestinationAddress common.Address `json:"dest_address"`
 	Amount             *big.Int       `json:"amount"`
+	MetadataIsHashed   bool           `json:"-"`
```
goran-ethernal:
```suggestion
	IsMetadataHashed   bool           `json:"-"`
```

ToniRamirezM (approved):

goran-ethernal (approved):

PR #235, opened 2024-12-09
fix: Aggsender high cpu usage

joanestebanr:
## Description

- Aggsender is using about 200% of CPU (version: `v0.5.0-beta8`)
- Add a warning if decrease blockNumber (from L1 provider)

### DIFF
--- a/aggsender/block_notifier_polling_test.go
+++ b/aggsender/block_notifier_polling_test.go
@@ -196,13 +196,13 @@ func newBlockNotifierPollingTestData(t *testing.T, config *ConfigBlockNotifierPo
 			CheckNewBlockInterval: 0,
 		}
 	}
-	EthClientMock := mocks.NewEthClient(t)
+	ethClientMock := mocks.NewEthClient(t)
 	logger := log.WithFields("test", "BlockNotifierPolling")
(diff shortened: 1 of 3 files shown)


goran-ethernal (approved):

aggsender/block_notifier_polling_test.go
```diff
@@ -193,7 +193,7 @@ func newBlockNotifierPollingTestData(t *testing.T, config *ConfigBlockNotifierPo
 	if config == nil {
 		config = &ConfigBlockNotifierPolling{
 			BlockFinalityType:     etherman.LatestBlock,
-			CheckNewBlockInterval: time.Second,
+			CheckNewBlockInterval: 0,
 		}
 	}
 	EthClientMock := mocks.NewEthClient(t)
```
Stefan-Ethernal: Nitpick (rename the other places in this scope as well, I couldn't add comment):
```suggestion
	ethClientMock := mocks.NewEthClient(t)
```

Stefan-Ethernal (approved):
We should consider putting the 80% scaling (https://github.com/0xPolygon/cdk/blob/dc335e444bed4995ca6202b3d40bf0d5c28577c8/aggsender/block_notifier_polling.go#L189-L190) into some const

PR #237, opened 2024-12-10
fix: calculate correct root for `LER` merkle proof and fix `BridgeExit` hash

goran-ethernal:
## Description

This PR fixes the issue of incorrect calculation of `BridgeExit` hash. Before this PR, we didn't use this hash anywhere in the aggsender, but now, it is needed to calculate the correct `LocalExitRoot` for the `ImportedBridgeExit` that is coming from L2 -> L2.

Before this fix, when a `ImportedBridgeExit` (`Claim`) comes from `L2 -> L2`, we were sending the wrong local exit tree root, and certificate with these claims always failed on the `agglayer`.

Now, `aggsender` calculates the correct local exit tree root when its a claim coming from a rollup, as in the `zkevm-contracts`: https://github.com/0xPolygonHermez/zkevm-contracts/blob/4912f4b673015209b3dbe1dd0702a9ffec5c9261/contracts/v2/PolygonZkEVMBridgeV2.sol#L808

### DIFF
--- a/aggsender/aggsender_test.go
+++ b/aggsender/aggsender_test.go
@@ -544,12 +544,12 @@ func TestGetImportedBridgeExits(t *testing.T) {
 				},
 			},
 		},
-		// {
-		// 	name:          "No claims",
-		// 	claims:        []bridgesync.Claim{},
-		// 	expectedError: false,
-		// 	expectedExits: []*agglayer.ImportedBridgeExit{},
-		// },
+		{
+			name:          "No claims",
+			claims:        []bridgesync.Claim{},
+			expectedError: false,
+			expectedExits: []*agglayer.ImportedBridgeExit{},
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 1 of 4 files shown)


joanestebanr (approved):

aggsender/aggsender_test.go
```diff
@@ -544,12 +544,12 @@ func TestGetImportedBridgeExits(t *testing.T) {
...
-			claims:        []bridgesync.Claim{},
-			expectedError: false,
-			expectedExits: []*agglayer.ImportedBridgeExit{},
-		},
+		// {
+		// 	name:          "No claims",
+		// 	claims:        []bridgesync.Claim{},
+		// 	expectedError: false,
+		// 	expectedExits: []*agglayer.ImportedBridgeExit{},
+		// },
```
Stefan-Ethernal: Uncomment or delete?

Stefan-Ethernal (approved):

aggsender/aggsender_test.go
goran-ethernal: Fixed: https://github.com/0xPolygon/cdk/pull/237/commits/c1f7c28cbad64458974af3d5d256025f27611511

PR #255, opened 2025-01-01
feat: mockery generate

temaniarpit27:
## Description

Mockery yaml file configuration

Fixes # (issue)

https://github.com/agglayer/aggkit/issues/73


.mockery.yaml
```diff
@@ -0,0 +1,145 @@
...
+      dir: "{{ .InterfaceDir }}/mocks"
+      filename: "mock_{{ .InterfaceName | snakecase | lower }}.go"
+      mockname: "{{ .InterfaceName }}Mock"
+    interfaces:
+      ProverInterface:
+      Etherman:
+      StorageInterface:
+      EthTxManagerClient:
+      RPCInterface:
+      Synchronizer:
```
temaniarpit27, line 28: @Stefan-Ethernal @goran-ethernal any idea how to generate with external package?
Couldn't find it and this doesn't work

.mockery.yaml
```diff
@@ -0,0 +1,149 @@
+with-expecter: true
+issue-845-fix: True
+resolve-type-alias: False
```
Stefan-Ethernal:
```suggestion
issue-845-fix: true
resolve-type-alias: false
```

.mockery.yaml
```diff
@@ -0,0 +1,149 @@
...
+dir: "{{ .InterfaceDir }}/../mocks"
+outpkg: "mocks"
+filename: "{{ .InterfaceName | snakecase | lower }}.go"
+mockname: "{{ .InterfaceName }}"
+packages:
+  github.com/0xPolygon/cdk/agglayer:
+    config:
+      inpackage: True
```
Stefan-Ethernal:
```suggestion
      inpackage: true
```

.mockery.yaml
```diff
@@ -0,0 +1,149 @@
...
+  github.com/0xPolygon/cdk/bridgesync:
+    config:
+      dir: "{{ .InterfaceDir }}/mocks"
+    interfaces:
+      ReorgDetector:
+      EthClienter:
+  github.com/0xPolygon/cdk/dataavailability:
+    config:
+      dir: "{{ .InterfaceDir }}/mocks_da"
+      all: True
```
Stefan-Ethernal:
```suggestion
      all: true
```

.mockery.yaml
```diff
@@ -0,0 +1,149 @@
...
+          - mockname: EthTxManagerMock
+            filename: mock_ethtxmanager.go
+            dir: "{{ .InterfaceDir }}/../test/helpers"
+            outpkg: "helpers"
+      Etherman:
+      RPCInterface:
+  github.com/0xPolygon/cdk/sequencesender/txbuilder:
+    config:
+      dir: "{{ .InterfaceDir }}/mocks_txbuilder"
+      all: True
```
Stefan-Ethernal:
```suggestion
      all: true
```

.mockery.yaml
```diff
@@ -0,0 +1,149 @@
+with-expecter: true
+issue-845-fix: True
+resolve-type-alias: False
+dir: "{{ .InterfaceDir }}/../mocks"
+outpkg: "mocks"
+filename: "{{ .InterfaceName | snakecase | lower }}.go"
```
Stefan-Ethernal: Let's have this pattern as default and then we can remove the override of this config in the specific packages. WDYT?
```suggestion
filename: "mock_{{ .InterfaceName | snakecase | lower }}.go"
```

.mockery.yaml
```diff
@@ -0,0 +1,149 @@
...
+    config:
+      dir: "{{ .InterfaceDir }}/mocks"
+      filename: "mock_{{ .InterfaceName | snakecase | lower }}.go"
+      mockname: "{{ .InterfaceName }}Mock"
+    interfaces:
+      ProverInterface:
+      Etherman:
+      StorageInterface:
+      EthTxManagerClient:
+      RPCInterface:
```
Stefan-Ethernal, line 34: I haven't been checking into details, but can we use the `all: true` for all the packages? Are there some interfaces for which we don't need the mocks?

.mockery.yaml
temaniarpit27: i think we can handle this where we will be standardizing everything

.mockery.yaml
temaniarpit27: yes, there were few which were generated which we dont need

.mockery.yaml
```diff
@@ -0,0 +1,149 @@
...
+filename: "{{ .InterfaceName | snakecase | lower }}.go"
+mockname: "{{ .InterfaceName }}"
+packages:
+  github.com/0xPolygon/cdk/agglayer:
+    config:
+      inpackage: true
+      dir: "{{ .InterfaceDir }}"
+      outpkg: "{{ .PackageName }}"
+    interfaces:
+      AgglayerClientInterface:
```
goran-ethernal, line 15: I see some interfaces that can be configured by the global mockery config, I don't think you need to add their specific config for them. Like this one. Maybe I am wrong 🤔

temaniarpit27: this cant be done here
to keep the filename same as it was earlier, we need to do specifically mention here. global config will generate - `agglayer_client_interface.go`
thats why i mentioned earlier that we need to standardize things
i would prefer to keep things as it is right now and pick up standardizing for all mocks in one go
let me know your thoughts.

goran-ethernal: Then I would say just standardize it with `mock_`. It doesn't matter if the file names are not the same. They are generated automatically, so no harm done.

temaniarpit27: done

Stefan-Ethernal (approved):

goran-ethernal (approved):

PR #265, opened 2025-01-14
fix: chainID in aggOracle configs

rachit77:
## Description

AggOracle should use `l2chainID` instead `l1chainID` in `etherman` configs since aggOracle is responsible for updating GERs in L2 smart contract.

Twin PR: https://github.com/agglayer/aggkit/pull/127


config/default.go
```diff
@@ -222,6 +222,8 @@ TargetChainType="EVM"
 URLRPCL1="{{L1URL}}"
 BlockFinality="FinalizedBlock"
 WaitPeriodNextGER="100ms"
+# L2ChainID is populated on runtime
+L2ChainID=0
```
Stefan-Ethernal: Do we need it, why we cannot use the `ChainIDL2` (https://github.com/0xPolygon/cdk/blob/abe8bf7697677041041b695c1311332209548a15/config/default.go#L230)?
```suggestion
```

aggoracle/config.go
```diff
@@ -21,5 +21,11 @@ type Config struct {
...
 	BlockFinality     string                   `jsonschema:"enum=LatestBlock, enum=SafeBlock, enum=PendingBlock, enum=FinalizedBlock, enum=EarliestBlock" mapstructure:"BlockFinality"` //nolint:lll
 	WaitPeriodNextGER types.Duration           `mapstructure:"WaitPeriodNextGER"`
+	L2ChainID         uint64                   `mapstructure:"L2ChainID"`
 	EVMSender         chaingersender.EVMConfig `mapstructure:"EVMSender"`
 }
+
+// ApplyL2ChainID copies L2ChainID to the L1ChainID in the Etherman config
+func (c *Config) ApplyL2ChainID(l2ChainID uint64) {
+	c.EVMSender.EthTxManager.Etherman.L1ChainID = l2ChainID
+}
```
Stefan-Ethernal: IMO, maybe it is worth considering renaming the `L1ChainID` config parameter simply to `ChainID` in the `etherman.Config` (https://github.com/0xPolygon/zkevm-ethtx-manager/blob/1b8de9df7e5b2d503d7b67753002cf06f036bebb/etherman/config.go#L15), set it explicitly and avoid this fuss. WDYT?

```suggestion
```

aggoracle/config.go
```diff
@@ -21,5 +21,11 @@ type Config struct {
 	// BlockFinality indicates the status of the blocks that will be queried in order to sync
 	BlockFinality     string                   `jsonschema:"enum=LatestBlock, enum=SafeBlock, enum=PendingBlock, enum=FinalizedBlock, enum=EarliestBlock" mapstructure:"BlockFinality"` //nolint:lll
 	WaitPeriodNextGER types.Duration           `mapstructure:"WaitPeriodNextGER"`
+	L2ChainID         uint64                   `mapstructure:"L2ChainID"`
```
Stefan-Ethernal: I'd say that we don't need this parameter, but instead use this [one](https://github.com/0xPolygon/cdk/blob/abe8bf7697677041041b695c1311332209548a15/aggoracle/chaingersender/evm.go#L53)? Or you want to read it from the `RollupManager` SC (which you are doing already)? We need to decide.

```suggestion
```

aggoracle/config.go
rachit77: We decided to abandon this idea because the transaction manager is widely used across various parts, and implementing this change would introduce an overwhelming number of adjustments to manage.

aggoracle/config.go
rachit77: should i remove [this](https://github.com/0xPolygon/cdk/blob/abe8bf7697677041041b695c1311332209548a15/aggoracle/chaingersender/evm.go#L53) param from config file and instead use `RollupManager` SC to populate this on runtime

Stefan-Ethernal: Yes for sure. But you don't need this one either (it is unused).

config/default.go
rachit77: removed  ref [3cd2d93](https://github.com/0xPolygon/cdk/pull/265/commits/3cd2d93209aefecdfbd65b6cf27899c165123f71)

aggoracle/config.go
rachit77: yeah removed it

aggoracle/config.go
```diff
@@ -23,3 +23,10 @@ type Config struct {
 	WaitPeriodNextGER types.Duration           `mapstructure:"WaitPeriodNextGER"`
 	EVMSender         chaingersender.EVMConfig `mapstructure:"EVMSender"`
 }
+
+// ApplyL2ChainID copies L2ChainID to relevant config fields
+func (c *Config) ApplyL2ChainID(l2ChainID uint64) {
+	// Ethereman needs to have the L2ChainID in the L1ChainID field for aggOacle
+	c.EVMSender.EthTxManager.Etherman.L1ChainID = l2ChainID
+	c.EVMSender.ChainIDL2 = l2ChainID
```
Stefan-Ethernal: Please correct me if I'm wrong, but the `ChainIDL2` is unused, so we can probably remove it as well (https://github.com/0xPolygon/cdk/blob/9634038b63f39151deefb1b38963ec487fdbd9ca/aggoracle/chaingersender/evm.go#L53)?

OTOH if we would need it at some point, with this approach that reads the l2 chain id from the rollup manager SC, it makes more sense to have that field exposed in the [EVMChainGERSender](https://github.com/0xPolygon/cdk/blob/9634038b63f39151deefb1b38963ec487fdbd9ca/aggoracle/chaingersender/evm.go#L59-L69), because it is not configurable (and therefore it could be misleading putting it in the configuration). WDYT?

rachit77: yup are correct `ChainIDL2` is unused as of now in aggOracle,
I will remove the `ChainIDL2` from `ApplyL2ChainID` method

Stefan-Ethernal (approved):

aggoracle/config.go
Stefan-Ethernal: Why don't we remove this field as well?
https://github.com/0xPolygon/cdk/blob/9634038b63f39151deefb1b38963ec487fdbd9ca/aggoracle/chaingersender/evm.go#L53

rachit77: Although that field isn't being used anywhere, I'm unsure about the intention behind adding it. Let me know if you'd like me to remove it.

rachit77: removed [d2f4050](https://github.com/0xPolygon/cdk/pull/265/commits/d2f40500cd77bf764b779aed249ef55dab1ef9e7)

joanestebanr (approved):

PR #267, opened 2025-01-15
fix: multi chains pessimistic proofs e2e tests

Stefan-Ethernal:
## Description

Bump the Kurtosis CDK version (https://github.com/0xPolygon/kurtosis-cdk/releases/tag/v0.2.26), because the transient branch, `jhilliard/multi-pp-testing` was merged in a meantime.

Fixes # (issue)

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -11,7 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-    - uses: actions/checkout@v4
+    - name: Checkout code
+      uses: actions/checkout@v4
+
     - uses: actions/setup-go@v5
       with:
         go-version: 1.22.x
@@ -44,9 +46,6 @@ jobs:
     steps:
     - name: Checkout Code
       uses: actions/checkout@v4
-
-    - name: Build Docker
-      run: make build-docker
     
     - name: Checkout Kurtosis CDK
       uses: actions/checkout@v4
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -53,7 +53,7 @@ jobs:
       with:
         repository: 0xPolygon/kurtosis-cdk
         path: kurtosis-cdk
-        ref: v0.2.26
+        ref: v0.2.25
     
     - name: Install Kurtosis CDK tools
       uses: ./kurtosis-cdk/.github/actions/setup-kurtosis-cdk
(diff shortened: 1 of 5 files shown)


.github/workflows/test-e2e.yml
```diff
@@ -42,24 +44,15 @@ jobs:
...
-      run: make build-docker
     
-    - name: Checkout kurtosis-cdk
+    - name: Checkout Kurtosis CDK
       uses: actions/checkout@v4
       with:
         repository: 0xPolygon/kurtosis-cdk
         path: kurtosis-cdk
-        ref: v0.2.24
+        ref: v0.2.25
```
rachit77, line 55: why not `v0.2.26` here as well?

rachit77 (approved):

.github/workflows/test-e2e.yml
Stefan-Ethernal: I have tried it, but then the agglayer [fails](https://github.com/0xPolygon/cdk/actions/runs/12804509856/job/35699332211) for some reason (I suspect that we would need to run with the newer image of the agglayer as well).

Maybe let's merge this one with using the v0.2.25 and then through a follow up PR we can experiment with the v0.2.26. WDYT?

rachit77: yup let's merge with `v0.2.25` for now

goran-ethernal (approved):

test/combinations/fork12-pessimistic-multi-attach-second-cdk.yml
```diff
@@ -27,7 +27,7 @@ args:
 
 
   cdk_node_image: cdk:latest
-  cdk_erigon_node_image: hermeznetwork/cdk-erigon:v2.60.0
+  cdk_erigon_node_image: hermeznetwork/cdk-erigon:v2.61.2
   zkevm_contracts_image: leovct/zkevm-contracts:v9.0.0-rc.3-pp-fork.12-patch.1
```
joanestebanr, line 31: I would fix bridge image  in both yml as in file `fork12-pessimistic.yml`

joanestebanr (approved):

test/combinations/fork12-pessimistic-multi-attach-second-cdk.yml
Stefan-Ethernal: Fixed https://github.com/0xPolygon/cdk/pull/267/commits/0ddd80b8f9074c6a5586857abced5131955defb8

vcastellm (approved):

PR #270, opened 2025-01-20
fix: seq-sender latestVirtualTime

### LINKED ISSUE #247: [BUG] Sequence-sender sends next sequence immediately after restart, ignoring LastBatchVirtualizationTimeMaxWaitPeriod configuration
# Bug Report

Sequence-sender sends next sequence immediately after restart, ignoring LastBatchVirtualizationTimeMaxWaitPeriod configuration.

## Description

We found a bug in the sequence-sender of the cdk-node service. The issue is with the timing of when the next sequences are sent. The configuration is set to send sequences every 3 hours, but when we restart the cdk-node, it sends the next sequence right away instead of waiting for the 3-hour period.

## To Reproduce

1. Wait for a batch to be created and then restart sequence-sender

## Actual behavior

Upon restarting the cdk-node, next sequences are sent immediately, bypassing the configured 3-hour interval.

## Expected behavior

Sequences should wait for the configured time period (3 hours) before being sent, even after restarting the cdk-node.

## Configuration

```
[SequenceSender]
MaxPendingTx = 5
SequencesTxFileName = "/tmp/sequencesender.json"
LastBatchVirtualizationTimeMaxWaitPeriod = "3h"
        [SequenceSender.EthTxManager]
                StoragePath = "/tmp/sequencesender-ethtxmanager.sqlite"
```

ToniRamirezM:
## Description

Gets latestVirtualTime from the ethtx manager db.

Fixes #247

### DIFF
--- a/sequencesender/ethtx.go
+++ b/sequencesender/ethtx.go
@@ -203,12 +203,13 @@ func (s *SequenceSender) syncEthTxResults(ctx context.Context) (uint64, error) {
 }
 
 // syncAllEthTxResults syncs all tx results from L1
-func (s *SequenceSender) syncAllEthTxResults(ctx context.Context) error {
+func (s *SequenceSender) syncAllEthTxResults(ctx context.Context) (time.Time, error) {
+	latestVirtualTime := time.Now().Local().Add(-s.cfg.LastBatchVirtualizationTimeMaxWaitPeriod.Duration)
 	// Get all results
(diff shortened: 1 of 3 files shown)


Stefan-Ethernal (approved):
LGTM

As a side note (unrelated to this PR, but caught my attention), maybe we should consider setting the UTC time for the `SentL1Timestamp` and `StatusTimestamp` fields...?

joanestebanr (approved):

PR #272, opened 2025-01-21
fix: `reorg detector` and `syncer` issues on `develop`

goran-ethernal:
## Description

This PR fixes the issues found on the `l1infotreesyncer` and `reorg detector`, when we might not be tracking a correct block hash in the `reorg detector`.

It also fixes an issue when download channel is closing, it passes an empty block object (since it is passing objects by value and not by reference), with number 0, and empty hash, which processor passes to the `reorg detector` which is not correct.

**Important Note**:
`e2e` tests use `sumulated client` as the `l1 client` node, which only knows of `SafeBlocks`. That is why this PR expanded the constructor function of the `EVMDownloader` to receive `finalizedBlockType` parameter. When we run the `cdk` node or any syncer that syncs data from `L1`, the `finalizedBlockType` is `FinalizedBlock` as seen in the `run.go`. `e2e` tests use `SafeBlock`.
Also, `L2` does not have a notion of `FinalizedBlock`, so `bridgeL2Syncer` will use the `LatestBlock` for this parameter.

Fixes # (issue)

### DIFF
--- a/sync/evmdownloader.go
+++ b/sync/evmdownloader.go
@@ -58,19 +58,27 @@ func NewEVMDownloader(
 	if err != nil {
 		return nil, err
 	}
+
 	topicsToQuery := make([]common.Hash, 0, len(appender))
 	for topic := range appender {
 		topicsToQuery = append(topicsToQuery, topic)
 	}
+
+	fbtEthermanType := finalizedBlockType
 	fbt, err := finalizedBlockType.ToBlockNum()
 	if err != nil {
 		return nil, err
 	}
+
 	if fbt.Cmp(finality) > 0 {
 		// if someone configured the syncer to query blocks by Safe or Finalized block
 		// finalized block type should be at least the same as the block finality
 		fbt = finality
+		fbtEthermanType = blockFinalityType
 	}
+
+	logger.Infof("downloader initialized with block finality: %s, finalized block type: %s. SyncChunkSize: %d",
+		blockFinalityType, fbtEthermanType, syncBlockChunkSize)
 
 	return &EVMDownloader{
 		syncBlockChunkSize: syncBlockChunkSize,
--- a/sync/evmdownloader.go
+++ b/sync/evmdownloader.go
@@ -39,7 +39,8 @@ type LogAppenderMap map[common.Hash]func(b *EVMBlock, l types.Log) error
 type EVMDownloader struct {
 	syncBlockChunkSize uint64
 	EVMDownloaderInterface
-	log *log.Logger
+	log                *log.Logger
+	finalizedBlockType etherman.BlockNumberFinality
 }
 
 func NewEVMDownloader(
@@ -83,6 +84,7 @@ func NewEVMDownloader(
 	return &EVMDownloader{
 		syncBlockChunkSize: syncBlockChunkSize,
 		log:                logger,
+		finalizedBlockType: fbtEthermanType,
 		EVMDownloaderInterface: &EVMDownloaderImplementation{
 			ethClient:              ethClient,
 			blockFinality:          finality,
@@ -136,19 +138,19 @@ func (d *EVMDownloader) Download(ctx context.Context, fromBlock uint64, download
 		blocks := d.GetEventsByBlockRange(ctx, fromBlock, toBlock)
 
 		if toBlock <= lastFinalizedBlockNumber {
-			d.reportBlocks(downloadedCh, blocks)
+			d.reportBlocks(downloadedCh, blocks, lastFinalizedBlockNumber)
 			fromBlock = toBlock + 1
 
 			if blocks.Len() == 0 || blocks[blocks.Len()-1].Num < toBlock {
-				d.reportEmptyBlock(ctx, downloadedCh, toBlock)
+				d.reportEmptyBlock(ctx, downloadedCh, toBlock, lastFinalizedBlockNumber)
 			}
 		} else {
-			d.reportBlocks(downloadedCh, blocks)
+			d.reportBlocks(downloadedCh, blocks, lastFinalizedBlockNumber)
 
 			if blocks.Len() == 0 {
 				if lastFinalizedBlockNumber > fromBlock &&
 					lastFinalizedBlockNumber-fromBlock > d.syncBlockChunkSize {
-					d.reportEmptyBlock(ctx, downloadedCh, fromBlock+d.syncBlockChunkSize)
+					d.reportEmptyBlock(ctx, downloadedCh, fromBlock+d.syncBlockChunkSize, lastFinalizedBlockNumber)
 					fromBlock += d.syncBlockChunkSize + 1
 				}
 			} else {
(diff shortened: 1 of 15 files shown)


sync/evmdownloader.go
```diff
@@ -86,28 +108,44 @@ func (d *EVMDownloader) Download(ctx context.Context, fromBlock uint64, download
...
+			d.log.Infof(
 				"waiting for new blocks, last block processed: %d, last block seen on L1: %d",
 				fromBlock-1, lastBlock,
 			)
 			lastBlock = d.WaitForNewBlocks(ctx, fromBlock-1)
 			continue
 		}
-		d.log.Debugf("getting events from block %d to %d", fromBlock, toBlock)
+
+		lastFinalizedBlock, err := d.GetLastFinalizedBlock(ctx)
```
joanestebanr, line 130: suggestion: instead of requesting lastFinalized use `BlockNotifier` interface from `aggsender` (move to a common place, for sure). And refactor also reorg that is doing the same query to RPC

sync/evmdownloader.go
```diff
@@ -86,28 +108,44 @@ func (d *EVMDownloader) Download(ctx context.Context, fromBlock uint64, download
...
+		lastFinalizedBlockNumber := lastFinalizedBlock.Number.Uint64()
+
+		d.log.Infof("getting events from blocks %d to  %d. lastFinalizedBlock: %d",
+			fromBlock, toBlock, lastFinalizedBlockNumber)
 		blocks := d.GetEventsByBlockRange(ctx, fromBlock, toBlock)
-		for _, b := range blocks {
-			d.log.Debugf("sending block %d to the driver (with events)", b.Num)
-			downloadedCh <- b
+
+		reportBlocksFn := func(numOfBlocksToReport int) {
```
joanestebanr: suggestion: create a function instead of declaring here

sync/evmdownloader.go
```diff
@@ -86,28 +108,44 @@ func (d *EVMDownloader) Download(ctx context.Context, fromBlock uint64, download
...
+			for i := 0; i < numOfBlocksToReport; i++ {
+				d.log.Infof("sending block %d to the driver (with events)", blocks[i].Num)
+				downloadedCh <- blocks[i]
+			}
 		}
-		if len(blocks) == 0 || blocks[len(blocks)-1].Num < toBlock {
+
+		reportEmptyBlockFn := func(blockNum uint64) {
```
joanestebanr: suggestion: create a function instead of declaring here

sync/evmdownloader.go
```diff
@@ -116,7 +154,21 @@ func (d *EVMDownloader) Download(ctx context.Context, fromBlock uint64, download
...
+
+		if toBlock <= lastFinalizedBlockNumber {
+			reportBlocksFn(blocks.Len())
+			fromBlock = toBlock + 1
+
+			if blocks.Len() == 0 || blocks[blocks.Len()-1].Num < toBlock {
+				reportEmptyBlockFn(toBlock)
+			}
+		} else {
+			lastFinalizedBlockNum, i, found := blocks.LastFinalizedBlock(lastFinalizedBlockNumber)
```
joanestebanr: The idea is to report the block that have logs even that are > `lastFinalizedBlockNumber`

- Ask logs for current range [ n - m ]
      -  Returns  log block x -> next_range = [  x - ( x+ `block_chunk`) ]
      - Returns empty: 
            -  if (  `fbn` - n > `block_chunck` ) -> create empty block with getHeader(fbn+block_chunk) 
                   - we have enough blocks on `finalized` zone to create a empty block  
                  - next_range =  [ (`fbn` + `block_chunck`)  - m + `block_chunck` ] 
             - else
                   - We just increase the range:
                  -  next_range = [ n - m + `block_chunck` ]

sync/evmdownloader.go
```diff
@@ -86,28 +108,44 @@ func (d *EVMDownloader) Download(ctx context.Context, fromBlock uint64, download
...
+				downloadedCh <- blocks[i]
+			}
 		}
-		if len(blocks) == 0 || blocks[len(blocks)-1].Num < toBlock {
+
+		reportEmptyBlockFn := func(blockNum uint64) {
 			// Indicate the last downloaded block if there are not events on it
 			d.log.Debugf("sending block %d to the driver (without events)", toBlock)
-			header, isCanceled := d.GetBlockHeader(ctx, toBlock)
+			header, isCanceled := d.GetBlockHeader(ctx, blockNum)
```
joanestebanr: the struct `EVMBlockHeader` need a new flag to say that is a 'safe block' and no need to be added to reorg detector

sync/evmdriver.go
joanestebanr, line 1: If the block is 'safe' enough don't need to add to reorg detector

joanestebanr (changes requested):

sync/evmdownloader.go
goran-ethernal: I would leave that for another PR since it is a big change, and it will require a lot of changes on the existing tests in the sync package.

sync/evmdownloader.go
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/272/commits/0982c9a37f3dbd873ec03752c886e9932aa1dbe1

sync/evmdownloader.go
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/272/commits/0982c9a37f3dbd873ec03752c886e9932aa1dbe1

sync/evmdownloader.go
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/272/commits/0982c9a37f3dbd873ec03752c886e9932aa1dbe1

sync/evmdriver.go
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/272/commits/dfadcb6d68281609ef21513466f71195079990c1

sync/evmdownloader.go
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/272/commits/dfadcb6d68281609ef21513466f71195079990c1

sync/evmtypes.go
```diff
@@ -2,14 +2,22 @@ package sync
...
+type EVMBlocks []*EVMBlock
+
+func (e EVMBlocks) Len() int {
+	return len(e)
+}
+
 type EVMBlock struct {
 	EVMBlockHeader
-	Events []interface{}
+	IsSafeBlock bool
```
joanestebanr: maybe the name is a bit missleading, because the block is `finalized` not safe

sync/evmdownloader.go
```diff
@@ -50,19 +52,39 @@ func NewEVMDownloader(
...
 	}
+
+	fbtEthermanType := finalizedBlockType
+	fbt, err := finalizedBlockType.ToBlockNum()
+	if err != nil {
+		return nil, err
+	}
+
+	if fbt.Cmp(finality) > 0 {
+		// if someone configured the syncer to query blocks by Safe or Finalized block
```
joanestebanr, line 75: I suggest to  emit a `warning` to mention that the configuration is override

sync/evmtypes.go
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/272/commits/81901ee97820069bdc769c0da86a31e408681b75

sync/evmdownloader.go
goran-ethernal: Fix: https://github.com/0xPolygon/cdk/pull/272/commits/81901ee97820069bdc769c0da86a31e408681b75

joanestebanr (approved):

ToniRamirezM (approved):

PR #275, opened 2025-01-23
fix: case of eth_getBlockByHash response null or tstamp zero

joanestebanr:
## Description

If RPC returns a `null` to the call `eth_getBlockByHash` then it get a tstamp of zero that produce a incorrect value for `maxSequenceTimestamp`

### DIFF
--- a/rpc/batch.go
+++ b/rpc/batch.go
@@ -116,7 +116,7 @@ func (b *BatchEndpoints) GetL2BlockTimestamp(blockHash string) (uint64, error) {
 	}
 	timestamp := new(big.Int).SetBytes(common.FromHex(l2Block.Timestamp)).Uint64()
 	if timestamp == 0 {
-		return 0, fmt.Errorf("error timestamp str:'%s' form Block hash: %s is 0", l2Block.Timestamp, blockHash)
+		return 0, fmt.Errorf("timestamp str '%s' from block hash %s is 0", l2Block.Timestamp, blockHash)
 	}
 	return timestamp, nil
 }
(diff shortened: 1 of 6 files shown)


rpc/batch.go
```diff
@@ -102,14 +103,22 @@ func (b *BatchEndpoints) GetL2BlockTimestamp(blockHash string) (uint64, error) {
...
 	l2Block := zkeEVML2Block{}
 	err = json.Unmarshal(response.Result, &l2Block)
 	if err != nil {
 		return 0, fmt.Errorf("error unmarshalling the l2 block from the response calling eth_getBlockByHash: %w", err)
 	}
-
-	return new(big.Int).SetBytes(common.FromHex(l2Block.Timestamp)).Uint64(), nil
+	timestamp := new(big.Int).SetBytes(common.FromHex(l2Block.Timestamp)).Uint64()
+	if timestamp == 0 {
+		return 0, fmt.Errorf("error timestamp str:'%s' form Block hash: %s is 0", l2Block.Timestamp, blockHash)
```
ToniRamirezM:
```suggestion
		return 0, fmt.Errorf("timestamp str '%s' from block hash %s is 0", l2Block.Timestamp, blockHash)
```

rpc/batch.go
```diff
@@ -102,14 +103,22 @@ func (b *BatchEndpoints) GetL2BlockTimestamp(blockHash string) (uint64, error) {
 		return 0, fmt.Errorf("error in the response calling eth_getBlockByHash: %v", response.Error)
 	}
 
+	if string(response.Result) == "null" {
+		log.Errorf("error response of eth_getBlockByHash  is null. Block hash: %s. err: Not Found", blockHash)
```
ToniRamirezM:
```suggestion
		log.Errorf("eth_getBlockByHash response is null. Block hash %s not found", blockHash)
```

ToniRamirezM (approved):

PR #277, opened 2025-01-23
fix: cherry-pick case of eth_getBlockByHash response null or tstamp zero (#275)

joanestebanr:
cherry-pick #275

### DIFF
--- a/rpc/batch.go
+++ b/rpc/batch.go
@@ -15,7 +15,8 @@ import (
 
 var (
 	// ErrBusy is returned when the witness server is busy
-	ErrBusy = errors.New("witness server is busy")
+	ErrBusy     = errors.New("witness server is busy")
+	jSONRPCCall = rpc.JSONRPCCall
 )
 
 const busyResponse = "busy"
@@ -92,7 +93,7 @@ func (b *BatchEndpoints) GetL2BlockTimestamp(blockHash string) (uint64, error) {
 
 	log.Infof("Getting l2 block timestamp from RPC. Block hash: %s", blockHash)
 
-	response, err := rpc.JSONRPCCall(b.url, "eth_getBlockByHash", blockHash, false)
+	response, err := jSONRPCCall(b.url, "eth_getBlockByHash", blockHash, false)
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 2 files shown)


rpc/batch.go
```diff
@@ -102,14 +103,22 @@ func (b *BatchEndpoints) GetL2BlockTimestamp(blockHash string) (uint64, error) {
 		return 0, fmt.Errorf("error in the response calling eth_getBlockByHash: %v", response.Error)
 	}
 
+	if string(response.Result) == "null" {
+		log.Errorf("eth_getBlockByHash response is null. Block hash %s not found", blockHash)
+		return 0, fmt.Errorf("error response of eth_getBlockByHash  is null. Block hash: %s. err: Not Found", blockHash)
```
Stefan-Ethernal, line 108:
```suggestion
		return 0, fmt.Errorf("error response of eth_getBlockByHash is null. Block hash: %s. err: Not Found", blockHash)
```

rpc/batch.go
```diff
@@ -102,14 +103,22 @@ func (b *BatchEndpoints) GetL2BlockTimestamp(blockHash string) (uint64, error) {
 		return 0, fmt.Errorf("error in the response calling eth_getBlockByHash: %v", response.Error)
 	}
 
+	if string(response.Result) == "null" {
```
Stefan-Ethernal, line 106: Would it make sense to check if the response.Result is an empty string, wdyt?

rpc/batch_test.go
```diff
@@ -14,8 +14,6 @@ import (
 )
 
 func Test_getBatchFromRPC(t *testing.T) {
-	t.Parallel()
```
Stefan-Ethernal, line 17: Why it was not ok running it in parallel? Was it some issue with the test, or it is the matter of easier debugging the reason why it is removed?

Stefan-Ethernal (approved):

rpc/batch_test.go
joanestebanr: In parallel the test fails because there are interactions between tests. It needs a refactor of the test to allow that

rpc/batch.go
joanestebanr: I don't think so, empiracally I test that this is the response for no block, and added an extra check that if tstamp==0 fails (that would cover the case of empty string)

rpc/batch.go
joanestebanr: I think that is better don't change it because it's a cherry-pick

ToniRamirezM (approved):

PR #285, opened 2025-01-24
fix: hash meddler error

goran-ethernal:
## Description

This PR fixes the issue when running `aggsender` with `zkevm-eth-txmanager`, where both of them have their own `HashMeddler` implementation. `eth-txmanager` was missing a logic for `HashMeddler` that `aggsender` needs.

Fix was introduced in the `zkevm-eth-txmanager` PR: https://github.com/0xPolygon/zkevm-ethtx-manager/pull/93. This PR updates the dependency and adds a new UT.

Fixes # (issue)


Stefan-Ethernal (approved):
Let's just tag the new version on the zkevm-ethtx-manager and reference it, but otherwise LGTM!

rachit77 (approved):

PR #286, opened 2025-01-24
chore: bring `develop` into `main`

Stefan-Ethernal:
## Description

Sync the `main` with the changes from `develop`

Fixes # (issue)


ToniRamirezM:
Shouldn't we move the contents of release/v0.4.2 to main instead of develop? I understand develop is beyond the release branches and next release branch should be taken from develop, but what needs to be sent to main is the content of the latest release.

Stefan-Ethernal:
> Shouldn't we move the contents of release/v0.4.2 to main instead of develop? I understand develop is beyond the release branches and next release branch should be taken from develop, but what needs to be sent to main is the content of the latest release.
> 
> 

Yes, realized that when discussed with @vcastellm. I have closed this one.

PR #297, opened 2025-02-11
test: Aggoracle 

rachit77:
## Description

e2e test for aggOracle


jhkimqd:
With https://github.com/0xPolygon/kurtosis-cdk/pull/474, we are attempting to squash together the sovereign rollup deployment and the multiple CDK deployments together. But since the AggKit doesn't rely on the existence of multiple CDKs, and can be tested with the OP rollup - I think this approach to isolate the sovereign rollup deployment makes more sense.

PR #310, opened 2025-03-14
fix: update Kurtosis CDK references and Docker images

Stefan-Ethernal:
## Description

Currently the e2e from branch release/v0.5.x are failing due kurtosis-cdk changes. This PR fix it.

The FEP and PP are splitted to reduce the complexity of pass both of them with same combination of agglayer + kurtosis_cdk

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -91,11 +91,11 @@ jobs:
 
     
     - name: Dump enclave logs
-      if: failure()
+      if: always()
       run: kurtosis dump ./dump
 
     - name: Generate archive name
-      if: failure()
+      if: always()
       run: |
         archive_name="dump_run_with_args_${{matrix.e2e-group}}_${{ github.run_id }}"
         echo "ARCHIVE_NAME=${archive_name}" >> "$GITHUB_ENV"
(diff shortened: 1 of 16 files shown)


.github/workflows/test-e2e.yml
```diff
@@ -91,19 +90,19 @@ jobs:
 
     
     - name: Dump enclave logs
-      if: failure()
+      if: always()
```
Stefan-Ethernal: I have enabled this in order to be able to have the logs (because the e2e tests were timing out). If you think we don't need it in any case, but only for the failures, we can revert back to `failure` trigger. WDYT?

test/run-e2e.sh
```diff
@@ -25,8 +26,8 @@ else
 fi
 
 kurtosis clean --all
-echo "Override cdk config file"
-cp $BASE_FOLDER/config/kurtosis-cdk-node-config.toml.template $KURTOSIS_FOLDER/templates/trusted-node/cdk-node-config.toml
+override_cdk_node_config_file  $DATA_AVAILABILITY_MODE
```
Stefan-Ethernal: Extra space

```suggestion
override_cdk_node_config_file $DATA_AVAILABILITY_MODE
```

ToniRamirezM (approved):

Stefan-Ethernal (commented):
LGTM, aside from nitpicks (I cannot approve it, since I have created the PR :melting_face:)

joanestebanr (approved):
@Stefan-Ethernal I can approve it!

Issue #325, opened 2025-05-01
[BUG] CDK Node Stopped Network Issue


praetoriansentry:
# Bug Report

During, some tests it looks like the CDK node lost connectivity to the L1 RPC and the CDK node itself exited with a fatal from this line:
https://github.com/0xPolygon/cdk/blob/b0bd30aef0ea1accd677949649a7328b5982540a/sequencesender/sequencesender.go#L369

From the code, why we're exiting instead of returning an error and retrying. Does this situation really require a fatal?

```log
 2025-05-01T06:25:40.256Z	[31mERROR[0m	sequencesender/sequencesender.go:524	error getting latest virtual batch, error: Post "http://el-1-geth-lighthouse:8545": EOF	{"pid": 26, "version": "v0.5.4-rc1", "module": "sequence-sender"}
 github.com/0xPolygon/cdk/sequencesender.(*SequenceSender).updateLatestVirtualBatch
 	sequencesender/sequencesender.go:524
 github.com/0xPolygon/cdk/sequencesender.(*SequenceSender).tryToSendSequence
 	sequencesender/sequencesender.go:367
 github.com/0xPolygon/cdk/sequencesender.(*SequenceSender).sequenceSending
 	sequencesender/sequencesender.go:249
 2025-05-01T06:25:40.256Z	[31mFATAL[0m	sequencesender/sequencesender.go:369	error getting latest sequenced batch, error: fail to get latest virtual batch	{"pid": 26, "version": "v0.5.4-rc1", "module": "sequence-sender"}
 github.com/0xPolygon/cdk/sequencesender.(*SequenceSender).tryToSendSequence
 	sequencesender/sequencesender.go:369
 github.com/0xPolygon/cdk/sequencesender.(*SequenceSender).sequenceSending
 	sequencesender/sequencesender.go:249

```

## Expected behavior

I would probably expect an error log rather than the container being restarted.

## Environment (please complete the following information):

- Version 0.5.4-rc1

PR #337, opened 2025-05-07
chore: fatalf to panicf

temaniarpit27:
#325


Stefan-Ethernal (commented):
I would consider following two points:
- would it make sense to terminate the aggregator in case there is an error in the sequence sender, using the shared context instance?
- add some reconnect logic, in case there is some issue with the RPC node?

I believe it was invoking `Fatalf`, because we wanted to terminate the node (in lacking of graceful shutdown of other components, such as aggregator, via shared context object which is definitely much better approach).

sequencesender/sequencesender.go
```diff
@@ -367,7 +368,8 @@ func (s *SequenceSender) tryToSendSequence(ctx context.Context) {
 	// Get latest virtual state batch from L1
 	err = s.updateLatestVirtualBatch()
 	if err != nil {
-		s.logger.Fatalf("error getting latest sequenced batch, error: %v", err)
+		s.logger.Errorf("error getting latest sequenced batch, error: %v", err)
```
temaniarpit27: @Stefan-Ethernal even if it return err and breaks, it will reconnect again right?

sequencesender/sequencesender.go
```diff
@@ -141,7 +141,8 @@ func (s *SequenceSender) Start(ctx context.Context) {
 	// Get latest virtual state batch from L1
 	err := s.updateLatestVirtualBatch()
 	if err != nil {
-		s.logger.Fatalf("error getting latest sequenced batch, error: %v", err)
+		s.logger.Errorf("error getting latest sequenced batch, error: %v", err)
```
temaniarpit27, line 144: @Stefan-Ethernal i think fatalf here makes sense as we dont want to stop node if sequence sender doesn't start.

sequencesender/sequencesender.go
temaniarpit27: errorf here makes sense to me

Stefan-Ethernal: In this case no, because there is no reconnect logic. The connecting is happenning [here](https://github.com/0xPolygon/cdk/blob/b8cb3ee32ab1942a396d029dbf176c9843e69cf9/etherman/etherman.go#L109-L114), when the Etherman is instantiated and it would reconnect only if we restart the node AFAICT.

joanestebanr (commented):
There are any prodution issue to change that? AFAIK the Fatalf are points with no recoverable errors, if there are no specific reason to change it I don't see any benefit.

temaniarpit27:
> There are any prodution issue to change that? AFAIK the Fatalf are points with no recoverable errors, if there are no specific reason to change it I don't see any benefit.

https://github.com/0xPolygon/cdk/issues/325
@praetoriansentry reported this

sequencesender/sequencesender.go
ToniRamirezM: If there is no reconnection or error handling I would leave the fatalf until the error handling is implemented.

temaniarpit27: decided not to handle the error @arnaubennassar @Stefan-Ethernal @joanestebanr @ToniRamirezM
just moved from fatalf to panicf

temaniarpit27: @praetoriansentry

praetoriansentry: > In this case no, because there is no reconnect logic. The connecting is happenning [here](https://github.com/0xPolygon/cdk/blob/b8cb3ee32ab1942a396d029dbf176c9843e69cf9/etherman/etherman.go#L109-L114), when the Etherman is instantiated and it would reconnect only if we restart the node AFAICT.

If you're using ethclient, the reconnection should be handled at a lower level. I.e you don't need to explicitly call `ethclient.Dial(cfg.EthermanConfig.URL)` again or kill the entire process. "The client reconnects automatically when the connection is lost."  https://pkg.go.dev/github.com/ethereum/go-ethereum@v1.15.11/rpc#DialOptions

Stefan-Ethernal (approved):

Issue #358, opened 2025-06-16
[BUG] some bridge events may be ignored when `MaxCertSize` is non-zero


xzav3r:
# Bug Report

## Description

When `MaxCertSize` is set to a non-zero value, some bridge events may be ignored. This happens because `toBlock` is updated in the `limitCertSize` method, but the outdated `toBlock` value is used when saving the certificate.

## To Reproduce
1. set `MaxCertSize = 1000000`
2. start a mock L1 and L2 and CDK
3. sending 1000 bridge txs on L1 and L2 and waiting the pp generated

## Expected behavior

no bridge events will be ignored

PR #359, opened 2025-06-16
fix: update toBlock and add CreatedAt when setting MaxCertSize

xzav3r:
1. Fixes [issue](https://github.com/0xPolygon/cdk/issues/358)
2. add missing field `CreatedAt` when `MaxCertSize` is set to a non-zero value
3. set `BlockNumber` at the beginning

### DIFF
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -270,7 +270,7 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayer.SignedCertif
 		NewLocalExitRoot:      certificate.NewLocalExitRoot,
 		PreviousLocalExitRoot: &prevLER,
 		FromBlock:             fromBlock,
-		ToBlock:               toBlock,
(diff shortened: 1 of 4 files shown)


arnaubennassar:
Hey, @xzav3r appreciate the work here, but we have moved the aggsender implementation to the [aggkit repo](https://github.com/agglayer/aggkit). The aggsender code is already removed from `develop`, and we will create a release for it soon.

Any contribution to aggsender should be done to AggKit, and if you are planning to use this component, please do it so from an AggKit release instead

xzav3r:
> Hey, @xzav3r appreciate the work here, but we have moved the aggsender implementation to the [aggkit repo](https://github.com/agglayer/aggkit). The aggsender code is already removed from `develop`, and we will create a release for it soon.
> 
> Any contribution to aggsender should be done to AggKit, and if you are planning to use this component, please do it so from an AggKit release instead

got it, thank you!

